PR #124, opened 2021-06-18
style: Confirmation Modal should include a label for the PIN input

staylorwr:
Screen Shot 2021-06-18 at 4 46 07 PM


staylorwr:
I think I prefer this w/ buttons that aren't stacked, but I'll defer to the existing designs or @caleboller

caleboller (approved):

PR #127, opened 2021-06-21
fix: Ensure most subscriptions are loaded before rendering the UI

staylorwr:
Quick fix for ensuring the UI has what it needs before rendering down the tree.


staylorwr:
Closing in favor of #128

PR #128, opened 2021-06-22
Send Token Transfer Messages

staylorwr:


src/views/Wallet/index.vue
```diff
@@ -447,22 +451,25 @@ const WalletIndex = defineComponent({
...
           cleanupTransactionSubs()
+          activeMessage.value = ''
           historyPagination.next({ size: PAGE_SIZE })
         }
       }))
     }
 
     // call transferTokens() with built options and subscribe to confirmation and results
-    const transferTokens = (transferTokensInput: TransferTokensInput, sc: TokenBalance) => {
+    const transferTokens = (transferTokensInput: TransferTokensInput, message: any, sc: TokenBalance) => {
```
Sajjon, line 468: you can change `message: any` to `message: MessageInTransaction` instead, might be helpful for debugging.

The type looks like this:

```typescript
export type MessageInTransaction = Readonly<{
	plaintext: string
	encrypt: boolean
}>
```

staylorwr: 👌🏼 Thanks for that, looking deeper now.

caleboller (approved):

PR #152, opened 2021-07-08
fix: Replace overflow-_-scroll with overflow-_-auto

staylorwr:
﻿https://www.loom.com/share/b0d0fd5455d0413b9201b42d82a59096


staylorwr:
@alexandreaco if you know of a better way to get that account nick name to correctly react I'm all ears, but was left with either refactoring the account nick name logic back up into `wallet/index` or the current solution of just adding an incremented name index so that the child components can watch for changes to hardware names.

PR #182, opened 2021-07-16
use high level method for unstakes

AbstractFruitFactory:
Depending on `radix.activeAddress()` to update unstakes does not work as expected, since the active address stream will only push new values when changing account (or when starting the app). 

Using the helper method `radix.unstakingPositions()` however, periodically pushes values and therefore will keep triggering the `subscribe` logic, updating the state as expected.

### DIFF
--- a/src/views/Wallet/index.vue
+++ b/src/views/Wallet/index.vue
@@ -281,11 +281,9 @@ const WalletIndex = defineComponent({
     subs.add(radix.accounts.subscribe((accountsRes: AccountsT) => { accounts.value = accountsRes }))
     subs.add(radix.activeAddress.subscribe((addressRes: AccountAddressT) => { activeAddress.value = addressRes }))
 
-    subs.add(radix.activeAddress.pipe(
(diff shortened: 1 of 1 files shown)


staylorwr (changes requested):
@AbstractFruitFactory Thanks for this PR.  I went back through our git history to look at why we made these changes and found #131.  My understanding is that this is still a problem with `unstakingPositions`,  ie:

> between epochs the radix.unstakingPositions endpoint is not sending pending unstakes and results in the UI showing less funds to the user. We need to switch to the ledger endpoint for now until this is fixed

I also found the original slack thread: https://townshipagency.slack.com/archives/C01PEGVLKJ8/p1623851923033800

I guess the ideal situation would be to remove the filter as you suggested at the bottom of the slack thread here: https://github.com/radixdlt/radixdlt-javascript/blob/main/packages/application/src/radix.ts#L377

AbstractFruitFactory:
Yep there was some issue back then - but changing this back now I'm seeing expected behavior so far. 

I will also remove the filter that we are using in the SDK, but theoretically that should not affect anything.

staylorwr (approved):

PR #197, opened 2021-07-20
feat: add proper verify address modals everywhere

caleboller:
This PR adds the verify address modal functionality to all relevant copy icons. When a hardware wallet is the active account, it will show the verify modal and cue the Ledger to verify the address before copying.

### DIFF
--- a/src/components/ClickToCopy.vue
+++ b/src/components/ClickToCopy.vue
@@ -11,12 +11,19 @@
 import { defineComponent } from 'vue'
 import { copyToClipboard } from '@/actions/vue/create-wallet'
 import { useToast } from 'vue-toastification'
+import { getHardwareWalletAddress } from '@/actions/vue/data-store'
 
 const ClickToCopy = defineComponent({
+
   props: {
-    text: {
+    address: {
       type: String,
       required: true
+    },
+    checkForHardwareAddress: {
+      type: Boolean,
+      required: false,
+      default: false
     }
   },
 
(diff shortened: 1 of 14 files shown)


src/components/ClickToCopy.vue
```diff
@@ -8,29 +8,40 @@
...
   props: {
-    text: {
+    address: {
       type: String,
       required: true
     }
   },
 
   setup () {
+    const hardwareAddress: Ref<string> = ref('')
```
staylorwr, line 27: This comes across as a code smell to me.  For nearly all cases, this component serves a single purpose of copying a string to the clipboard.   

With these new changes, the copy to clipboard component is now responsible for _always_ requesting the hardware address through the ipc, and then making it's own decisions about whether to emit an event or perform it's named responsibility.  

I'd advocate for either:

- writing a separate component named something like 'VerifyAndCopy' that contains this functionality, leaving the 'ClickToCopy' to handle other types of addresses
- add a new prop to this component that defines which function the component serves, ie:
```
props: {
  address: {
    type: String,
    required: true
  },
  action: {
    type: String,
    required: false,
    default: 'copy'
   }
}
```
In cases that you might be verifying, declare it as such w/ a `<click-to-copy action="verify"...>` so that it's clear what the intent of the component is.

staylorwr (changes requested):

src/components/ClickToCopy.vue
caleboller: @staylorwr I've cleaned this up quite a bit and implemented one of your advised solutions:

- an optional `checkForHardwareAddress` can be sent to the component, defaulting to false.
- If that above option is set to true, the component will (at click time), fetch the hardware address from disk and compare, and then take the appropriate action - either instantly copying or showing the verify modal.

This implementation ensures we're not reading too frequently from disk, and makes the component friendly to use in list views. I should have caught this beforehand.

staylorwr (approved):

PR #216, opened 2021-07-23
feat: New PIN entry

alexandreaco:
Screen Shot 2021-07-23 at 2 43 45 PM
Screen Shot 2021-07-23 at 2 43 49 PM
Screen Shot 2021-07-23 at 2 10 20 PM
Screen Shot 2021-07-23 at 2 12 40 PM


staylorwr:
Sure looks nice to me!

PR #229, opened 2021-07-26
Chore: Revent hardware connection events & enforce certs

caleboller:
This PR includes two changes:

1) I've reverted the most recent round of connection event updates implemented by @AbstractFruitFactory. They're incredible, but are causing crash logs to be shown no quit for non-Ledger users. We're going to return those post mainnet release.
2) Remove our early implemented solutions to ignore cert issues. For mainnet release, we need to require that all nodes have valid certs.


staylorwr (approved):
👍🏼

PR #240, opened 2021-07-27
Chore: Remove all the testnet

caleboller:
This PR removes the faucet button and the last remaining testnet copy.


staylorwr (approved):
👋🏼

PR #266, opened 2021-07-29
chore: improve how we handle network selection

caleboller:
This PR improves how we manage network selection in environment variables, and should help improve our Github build process long term.

We now expect an `VUE_APP_NETWORK_NAME` environment variable. Currently, we only support values of `stokenet` or `mainnet`. If we don't receive one of those two values, we throw an error. Choosing a network implictly has burned us in the past, I'd prefer we just bomb out.

### DIFF
--- a/.github/workflows/release-ledger.yml
+++ b/.github/workflows/release-ledger.yml
@@ -20,26 +20,11 @@ jobs:
           node-version: '15'
           check-latest: true
 
-      - name: Install node-hid Dependencies
-        shell: bash
-        run: |
-          case $RUNNER_OS in
-            Linux)
-              echo "Building node-hid deps for Ubuntu"
-              sudo apt install build-essential git
-              sudo apt install gcc-4.8 g++-4.8 && export CXX=g++-4.8
-              sudo apt install libusb-1.0-0 libusb-1.0-0-dev
-              sudo apt install libudev-dev
-              ;;
-            *)
-              echo "No commands required for this OS"
-          esac
       - name: Build/release Electron app
         uses: samuelmeuli/action-electron-builder@v1
         env:
-          VUE_APP_API: https://sandpitnet.radixdlt.com
-          VUE_APP_EXPLORER: https://sandpitnet-explorer.radixdlt.com
-          VUE_APP_FAUCET: https://sandpitnet-faucet.radixdlt.com/faucet/request
+          VUE_APP_NETWORK_NAME: mainnet
+          VUE_APP_EXPLORER: https://explorer.radixdlt.com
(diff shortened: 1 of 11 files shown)


src/views/CreateWallet/index.vue
```diff
@@ -113,10 +114,11 @@ const CreateWallet = defineComponent({
     const mnemonic: MnemomicT = Mnemonic.generateNew()
     const step = ref(0)
     const passcode = ref('')
+    const activeNetwork = network(process.env.VUE_APP_NETWORK_NAME)
```
staylorwr, line 117: 🤔 You're always passing in the environment variable to the `network` function with the same variable name.  Would it be better to just call the env variable in the `network` function?  I think you could also return a full wallet / sdk:

Instead of:
```ts
const activeNetwork = network(process.env.VUE_APP_NETWORK_NAME)
const radix = Radix
    .create({ network: Network.MAINNET })
    .connect(process.env.VUE_APP_API || 'https://mainnet.radixdlt.com')
```

maybe:
```ts
export const network = (networkName: string | undefined): ChosenNetworkT => {
  let response: ChosenNetworkT
  if (networkName === 'stokenet') {
    response = {
      network: Network.STOKENET,
      networkURL: 'https://stokenet.radixdlt.com'
    }
  } else if (networkName === 'mainnet') {
    response = {
      network: Network.MAINNET,
      networkURL: 'https://mainnet.radixdlt.com'
    }
  } else {
    throw new Error('Invalid Network Name Provided')
  }

  return response
}

export const newClient():  {
  const activeNetwork = network(process.env.VUE_APP_NETWORK_NAME)
  return Radix.create({ network: Network.MAINNET })
    .connect(process.env.VUE_APP_API || 'https://mainnet.radixdlt.com')
}

const radix = newClient().withWallet(store.state.wallet)
```

staylorwr (commented):
Thanks for these updates @caleboller, just a small opportunity to further clean this up, not a deal breaker to me though.

src/views/CreateWallet/index.vue
caleboller: I like 'dis. Will do.

.github/workflows/release-ledger.yml
```diff
@@ -20,26 +20,11 @@ jobs:
           node-version: '15'
           check-latest: true
 
-      - name: Install node-hid Dependencies
```
staylorwr, line 23: @caleboller We should update the branch to `main` and rename the workflow

staylorwr (changes requested):
A few updates on this workflow for latest builds.

staylorwr (approved):
👌🏼

PR #267, opened 2021-07-30
fix: validate network addresses for active network

caleboller:
This PR introduces a lightweight check to verify that user input addresses can both be safely unwrapped and also have valid preambles (`rdx` for mainnet, `tdx` for stokenet, etc) for the wallet's current network.


alexandreaco (approved):
Nice change!

AbstractFruitFactory (approved):

PR #268, opened 2021-08-02
Feat: The confirmation modal should submit on Enter, and Cancel on Escape.

caleboller:
This PR contains one fix and one new feature:

- The confirmation modal form previously contained two `button` elements. This was preventing the form from submitting successfully and closing the modal. Changing the cancel `button` to an anchor tag corrected the issue.

- Adds functionality to close the confirmation modal when the escape key is pressed.


staylorwr (approved):
💯 works well locally

PR #269, opened 2021-08-02
fix: Remove error handling for the completion event

staylorwr:
This PR cleans up duplicate, conflicting error handling that caused a regression when created an extraneous error message after reconnecting to a disconnected ledger.

Repro steps from the linear issue:

* Launch wallet, connect/PIN in/launch app on ledger, select hardware wallet to derive
* Disconnect ledger
* Go to send screen and enter a transaction, get to confirmation screen, click confirm
* Ledger prompt modal appears, telling me I need to retry since ledger is not connected
* Connect/PIN in/launch app on ledger, click retry
* Modal disappears and tosses me back to transaction form with error, transaction is not submitted

This PR simplifies the error behavior (I think all events always includes a completion error, but that's an assumption on my part, I've not been able to find the specific implementation in the SDK to validate that.)

Because of the structure of the transaction flow, I'm not sure that it's currently possible to automatically submit the transaction once the ledger, wallet and account have all been rederived and connected.  I guess that we could track the `activeTransaction = Ref<TransactionTracking | null>` in the Wallet index and check to see if it exists when the `connectHardwareWallet` callback is completed, but there's so much state here that I'm unsure if that would do what we want.  Any advice here @alexandreaco on submitting the transaction once the `connectHardwareWallet` function is complete? This PR currently clears the extra error out of the confirmation modal, allowing the user to click submit again.

### DIFF
--- a/src/views/Wallet/index.vue
+++ b/src/views/Wallet/index.vue
@@ -247,6 +247,7 @@ const WalletIndex = defineComponent({
     const activeAddress: Ref<AccountAddressT | null> = ref(null)
     const activeStakes: Ref<StakePositions | null> = ref(null)
     const activeUnstakes: Ref<UnstakePositions | null> = ref(null)
+    const activeTransaction: Ref<TransactionTracking | null> = ref(null)
     const accounts: Ref<AccountsT | null> = ref(null)
     const tokenBalances: Ref<TokenBalances | null> = ref(null)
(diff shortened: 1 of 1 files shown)


alexandreaco (approved):
I don't have my ledger setup to test this, but removing the extra error logic sounds like the right solution. While you're in there, can you also update the comment on line 506 to describe the new behavior?

PR #271, opened 2021-08-04
feat: On suspend, stop polling and send user home

caleboller:
This PR adds a small feature: When electron detects the host device is going into a suspended state (sleep), we send users currently viewing their wallet home, requiring them to enter their password to re-enter the wallet when the host device comes out of the suspended state.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,7 +2,14 @@
   <div class="relative bg-gradient-to-r from-rBlue via-blueMid to-blueEnd min-h-screen overflow-hidden font-sans antialiased">
     <div class="bg-rBlue w-1/2 h-tallest origin-center transform rotate-45 absolute z-0 top-0 left-0 -translate-x-1/2 -translate-y-1/2"></div>
     <div class="relative z-10 min-h-screen">
-      <router-view/>
+      <suspense>
+        <template #default>
+          <router-view/>
+        </template>
+        <template #fallback>
+          <h1> Loading </h1>
+        </template>
+      </suspense>
     </div>
   </div>
 </template>
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,7 +7,13 @@
           <router-view/>
         </template>
         <template #fallback>
-          <h1> Loading </h1>
+          <div class="bg-white h-full flex flex-col flex-1 w-full justify-around box-border">
+            <svg width="80" height="80" viewBox="0 0 80 80" fill="none" xmlns="http://www.w3.org/2000/svg" class="container animate-spin">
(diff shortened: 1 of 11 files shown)


src/App.vue
```diff
@@ -2,7 +2,14 @@
   <div class="relative bg-gradient-to-r from-rBlue via-blueMid to-blueEnd min-h-screen overflow-hidden font-sans antialiased">
     <div class="bg-rBlue w-1/2 h-tallest origin-center transform rotate-45 absolute z-0 top-0 left-0 -translate-x-1/2 -translate-y-1/2"></div>
     <div class="relative z-10 min-h-screen">
-      <router-view/>
+      <suspense>
```
saminakh, line 5: https://vueschool.io/articles/vuejs-tutorials/suspense-new-feature-in-vue-3/ reading about some of the suspense stuff here, this is a pretty nice pattern!

src/background.ts
```diff
@@ -81,6 +83,11 @@ app.on('activate', () => {
 // initialization and is ready to create browser windows.
 // Some APIs can only be used after this event occurs.
 app.on('ready', async () => {
+
+  electron.powerMonitor.on('suspend', () => {
+    win.webContents.send('resetToHome')
```
saminakh, line 88: 👍

saminakh (commented):
These changes are neat! And good teaching material ✨ Was the async connection changes required to reconnect upon 'waking up'?

src/App.vue
caleboller: This was my first time using it as it's a recent addition to Vue 3! Pretty nice indeed!

src/background.ts
caleboller: This `web.contents.send` function is how we send messages from the `main` process to the `render` process, you'll see elsewhere we use `ipcRenderer` to send messages from the `render` process to `main`.

src/background.ts
```diff
@@ -18,8 +18,10 @@ import {
 } from './actions/electron/data-store'
 import { sendAPDU } from './actions/electron/hardware-wallet'
 import menu from './menu'
+import electron from 'electron'
 
 const isDevelopment = process.env.NODE_ENV !== 'production'
+let win: BrowserWindow
```
alexandreaco, line 24: Why did you need to declare this up top as a `let`?

src/App.vue
```diff
@@ -2,7 +2,14 @@
...
+        <template #default>
+          <router-view/>
+        </template>
+        <template #fallback>
+          <h1> Loading </h1>
```
alexandreaco, line 10: Should we add the real loading icon here so we don't show an unstyled view to the end user?

alexandreaco (changes requested):
Only blocking change here is to replace the "loading" string with the loading svg

staylorwr (approved):
I tested a fresh build of this locally and it works as expected 💯

src/background.ts
caleboller: @alexandreaco it's being defined in a function and thus previously was being scoped to that function. I need later access to it in a event callback, so I had to declare it up top.

alexandreaco (approved):
Thanks for adding the loader 👍

PR #273, opened 2021-08-04
feat: add tooltip text upon hovering over Validator Name

SK-Sam:
This PR warns the user about the Validator URL and displays the URL upon hovering its name via tooltip.

[Demo of Tooltip](https://www.loom.com/share/12f43902f9634d5b8bd4d8df7db0ba99)

### DIFF
--- a/src/components/Tooltip.vue
+++ b/src/components/Tooltip.vue
@@ -1,0 +1,20 @@
+<template>
+  <div class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 rounded-sm shadow border border-solid border-rGrayLight w-60">
+    {{ descriptionText }}
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+
+const Tooltip = defineComponent({
+  props: {
+    descriptionText: {
+      type: String,
+      required: true
+    }
+  }
+})
+
+export default Tooltip
+</script>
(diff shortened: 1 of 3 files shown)


saminakh (commented):
These changes LGTM, and are working locally! I'm not sure I'm qualified yet to give a real approval though 🤔

alexandreaco (dismissed):
This tooltip works great! Let's turn this into a reusable component so it's a quicker addition next time. Happy to pair and show you what I mean :)

saminakh (commented):
these changes are looking good to me! ✨

src/components/Tooltip.vue
```diff
@@ -0,0 +1,20 @@
...
+  <div class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 rounded-sm shadow border border-solid border-rGrayLight w-60">
+    {{ descriptionText }}
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue'
+
+const Tooltip = defineComponent({
+  props: {
```
caleboller, line 11: @SK-Sam Instead of using a prop here, you might want to consider using the [slot](https://vuejs.org/v2/guide/components-slots.html) instead. This would allow us to render not only strings in the tooltip component, but any html we like.

caleboller (changes requested):

src/components/Tooltip.vue
saminakh: Is there a scenario where we'd want to put something other than a string in a tooltip? I suppose maybe being able to bold or italicize something with those html tags, is that the kind of scenario you were thinking of?

SK-Sam: @caleboller cool feature, TIL. Added `slots` and removed `props` 😄

staylorwr (approved):

PR #281, opened 2021-08-12
chore: Show error when new hw address added

caleboller:
This PR adds functionality to present users with an error modal when they attempt to derive a new Radix address with one already loaded. In the current build, this is only possible by using multiple ledger devices. Users are required instead to explicitly remove a hardware wallet before adding another.

It also adds a reusable error component for future use. :)


staylorwr (approved):
Code looks fine to me! 🆗 thanks for the little component!

PR #284, opened 2021-08-19
fix: Regressions in password error messages and forgot password submission

staylorwr:
This PR fixes:

### [Regression: No error on incorrect password](https://linear.app/township/issue/RDX-222/regression-no-error-on-incorrect-password)

> At some point entering an incorrect password stopped showing an error message.

At some point in time, we removed the Error message component from the home page.

and

### [Regression: "I forgot my password" link no longer words](https://linear.app/township/issue/RDX-223/regression-i-forgot-my-password-link-no-longer-works)

> At least not reliably. Some users manage to get it to work by "clicking around", but I confirmed that in general it's just a dead link.

The home routes 'modal' was driven by a prop that once initialized was not tracked.  This PR fixes that by introducing a `ref` for the value rather than relying on the initial prop.

### DIFF
--- a/src/views/Home/HomeEnterPasscode.vue
+++ b/src/views/Home/HomeEnterPasscode.vue
@@ -17,6 +17,7 @@
           data-ci="create-wallet-passcode-input"
           id="password"
           :validateOnInput="false"
+          :validateOnBlur="false"
         />
         <FormErrorMessage name="password" class="text-sm text-red-400" />
       </div>
--- a/src/views/Home/HomeEnterPasscode.vue
+++ b/src/views/Home/HomeEnterPasscode.vue
@@ -16,17 +16,18 @@
           rules="required"
           data-ci="create-wallet-passcode-input"
           id="password"
-          :validateOnInput="true"
+          :validateOnInput="false"
         />
+        <FormErrorMessage name="password" class="text-sm text-red-400" />
       </div>
 
       <ButtonSubmit class="w-96" :disabled="disableSubmit">
         {{ $t('home.passwordButton') }}
       </ButtonSubmit>
-      <div>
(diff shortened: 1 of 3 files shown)


src/views/Home/HomeEnterPasscode.vue
```diff
@@ -68,7 +72,14 @@ const HomeEnterPasscode = defineComponent({
     if (passEl) passEl.focus()
   },
 
-  emits: ['submit', 'forgotPassword']
+  emits: ['submit', 'forgotPassword'],
+
+  methods: {
+    forgotPassword () {
+      console.log('help')
```
saminakh, line 79: Should this still be in here?

saminakh (approved):
These changes LGTM! I tested them locally and both are working for me. Left one little note on a maybe stray console.log?

src/views/Home/index.vue
```diff
@@ -77,13 +77,19 @@ const CreateWallet = defineComponent({
...
+      if (props.modal) { uiModal.value = props.modal }
+    })
+
     const hasWallet = reactive({ value: null as boolean | null })
     const store = useStore()
     const router = useRouter()
     const { t } = useI18n({ useScope: 'global' })
 
-    const enterPasscodeComponent = ref(null)
+    const enterPasscodeComponent = rxRef(null)
```
SK-Sam, line 92: 👍

SK-Sam (commented):
Great, went through password error/forgot. 

Right now if i enter an incorrect password, `Open Wallet` button grays out and sends an error into Console.
If the message is supposed to be only in Console and not a popup/red text/alert then this is working fine!

src/views/Home/HomeEnterPasscode.vue
staylorwr: nope. 🙈

PR #286, opened 2021-08-19
fix: Show Word Numbers when revealing the seed phrase in settings

staylorwr:
Screen Shot 2021-08-19 at 1 30 25 PM


saminakh (approved):
LGTM! 👍

PR #288, opened 2021-08-20
chore: update styling of Wallet Overview page

SK-Sam:
This PR changes the styling so the other token balance arrow svg matches token name spacing

Screen Shot 2021-08-20 at 1 14 41 PM


saminakh (approved):
👍

PR #291, opened 2021-08-26
chore: Enable copying and pasting txn messages

alexandreaco:
Screen Shot 2021-10-21 at 11 35 59 AM


saminakh:
I can definitely see @caleboller 's point, but i also feel that maybe disabling everything is heavy-handed. Maybe there are certain components where it'd make sense to re-enable user select, or just enable on a case-by-case basis?

PR #293, opened 2021-09-02
limit retries to txhistory call

AbstractFruitFactory:


staylorwr:
Closing in favor of #296 and the networking work.  Thanks for figuring this out @AbstractFruitFactory

PR #294, opened 2021-09-03
chore: temporarily, latest builds target stokenet

caleboller:
There's a need to deliver stokenet builds for testing purposes. We'd like those stokenet builds to be produced as similarly as possible to our mainnet builds. So, temporarily, all builds on the latest channel will target stokenet.

This work will be undone when the imminent Network Selection work is complete.

### DIFF
--- a/.github/workflows/release-latest.yml
+++ b/.github/workflows/release-latest.yml
@@ -23,8 +23,8 @@ jobs:
       - name: Build/release Electron app
         uses: samuelmeuli/action-electron-builder@v1
         env:
-          VUE_APP_NETWORK_NAME: mainnet
-          VUE_APP_EXPLORER: https://explorer.radixdlt.com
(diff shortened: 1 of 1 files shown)


SK-Sam (approved):

.github/workflows/release-latest.yml
```diff
@@ -23,8 +23,8 @@ jobs:
       - name: Build/release Electron app
         uses: samuelmeuli/action-electron-builder@v1
         env:
-          VUE_APP_NETWORK_NAME: mainnet
-          VUE_APP_EXPLORER: https://explorer.radixdlt.com
+          VUE_APP_NETWORK_NAME: stokenet
+          VUE_APP_EXPLORER: https://stokenet-explorer.radixdlt.com
```
SK-Sam, line 27: Stoked!

PR #295, opened 2021-09-09
chore: Migrate store structure to handle multiple network options

alexandreaco:
Here is the migration


staylorwr (approved):
👌🏼 looks good to me.

PR #296, opened 2021-09-16
feat: Introduce NativeToken and TokenBalances composables

staylorwr:
Now that we've done all the hard work of moving _all_ of the logic from the `Wallet/index.vue` component into a few composables (is that the right word?), let's take it one step forward.

This PR separates out two major concerns of the `useWallet` composition api into separate, independent composables.

Doing this allows for each _route_ in the view tree to handle it's own subscriptions.  Each route that needs the token balances or native token associated with the authenticated wallet, can subscribe to that data independently, and then _unsubscribe_ when it's done.  This simplifies the API, increases separation of concerns, and also completely solves the problem of the `WalletOverview` and other views breaking on network change.

We should do this for:
- [x] TransactionHistory
- [x] Staking
- [x] Unstaking

If we continue down this path, this also means that each wallet will only subscribe to the data it needs to display _right now_, rather than all subs all the time.

There are probably some places contained in these changes that created regressions due to the inter-dependent nature of some of these subscriptions that we should clean up.  (native tokens in staking, transaction workflow comes to mind as well as account switching).

I think given the complexity of the transaction workflow, it might make sense to leave that contained in the `useWallet` composable for now, but we should validate that expectation.

### DIFF
--- a/src/composables/useTokenBalances.ts
+++ b/src/composables/useTokenBalances.ts
@@ -1,0 +1,27 @@
+import { ref, Ref } from 'vue'
+import { RadixT, Token, TokenBalance, TokenBalances } from '@radixdlt/application'
+
+interface useTokenBalancesInterface {
+  readonly tokenBalances: Ref<TokenBalances | null>;
+  tokenBalancesUnsub: () => void;
+  tokenBalanceFor: (token: Token) => TokenBalance | null;
+}
+
+export default function useTokenBalances (radix: RadixT): useTokenBalancesInterface {
+  const tokenBalances: Ref<TokenBalances | null> = ref(null)
+  const tokenBalancesSub = radix.tokenBalances.subscribe((tokenBalancesRes: TokenBalances) => {
+    tokenBalances.value = tokenBalancesRes
+  })
+
+  return {
+    tokenBalances,
+    tokenBalancesUnsub: () => {
+      tokenBalancesSub.unsubscribe()
+    },
+
+    tokenBalanceFor: (token: Token) => {
+      if (!tokenBalances.value) return null
+      return tokenBalances.value.tokenBalances.find((tb: TokenBalance) => tb.token.rri.equals(token.rri)) || null
+    }
+  }
+}
--- a/src/composables/useWallet.ts
+++ b/src/composables/useWallet.ts
@@ -18,11 +18,8 @@ import {
   SigningKeychain,
   SigningKeychainT,
   StakeOptions,
-  StakePositions,
   StakeTokensInput,
-  Token,
   TokenBalance,
-  TokenBalances,
   TransactionHistory,
   TransactionHistoryOfKnownAddressRequestInput,
   TransactionIdentifierT,
@@ -34,7 +31,6 @@ import {
   TransferTokensInput,
   TransferTokensOptions,
   UnstakeOptions,
-  UnstakePositions,
   UnstakeTokensInput,
   WalletErrorCause,
   WalletT,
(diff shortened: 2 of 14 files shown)


src/composables/useTokenBalances.ts
```diff
@@ -0,0 +1,27 @@
...
+    tokenBalances.value = tokenBalancesRes
+  })
+
+  return {
+    tokenBalances,
+    tokenBalancesUnsub: () => {
+      tokenBalancesSub.unsubscribe()
+    },
+
+    tokenBalanceFor: (token: Token) => {
```
alexandreaco, line 22: Can you add the return type of `: TokenBalance | null`?

src/composables/useWallet.ts
```diff
@@ -237,38 +231,18 @@ export default function useWallet (radix: RadixT, router: Router): useWalletInte
...
-    radix.ledger.nativeToken()
+    radix.unstakingPositions
   )
 
   const waitUntilAllLoaded = async () => firstValueFrom(allLoadedObservable)
 
   const reloadSubscriptions = () => reloadTrigger.next(Math.random())
 
   const initWallet = (): void => {
-    subs.add(reloadTrigger.asObservable()
```
alexandreaco, line 254: 👍

src/composables/useTokenBalances.ts
```diff
@@ -0,0 +1,27 @@
...
+
+interface useTokenBalancesInterface {
+  readonly tokenBalances: Ref<TokenBalances | null>;
+  tokenBalancesUnsub: () => void;
+  tokenBalanceFor: (token: Token) => TokenBalance | null;
+}
+
+export default function useTokenBalances (radix: RadixT): useTokenBalancesInterface {
+  const tokenBalances: Ref<TokenBalances | null> = ref(null)
+  const tokenBalancesSub = radix.tokenBalances.subscribe((tokenBalancesRes: TokenBalances) => {
```
alexandreaco, line 12: Does this subscription poll for the user's token balance? If not, we may want to add a refresh button on the Overview and Transaction pages that reloads the route if a user is expecting new tokens to appear

src/views/Wallet/WalletStaking.vue
```diff
@@ -216,33 +224,32 @@ const WalletStaking = defineComponent({
     }
 
     const handleSubmitStake = () => {
-      if (meta.value.valid && nativeTokenBalance.value) {
-        const safeAddress = safelyUnwrapValidator(values.validator)
-        const safeAmount = safelyUnwrapAmount(Number(values.amount))
-        const greaterThanZero = safeAmount && validateGreaterThanZero(safeAmount)
-        const validAmount = safeAmount && validateAmountOfType(safeAmount, nativeTokenBalance.value.token)
+      if (!tokenBalances.value || !nativeToken.value) return
```
alexandreaco, line 227: We're losing the `if(meta.value.valid)` guard here

alexandreaco (approved):
This looks like a big improvement. The only thing that really should be fixed is the missing check on a valid form submission within WalletStaking

src/views/Wallet/WalletStaking.vue
staylorwr: https://github.com/radixdlt/olympia-wallet/pull/296/files/133bc95a748cc2b36ecc2572486a1bcdc1052512#diff-e69d8164d1f3ddcc0f282d78fe249d369e1b040dafbc39b40bd3c8e8f4cabf1aR229
it's two lines down 😄

src/composables/useTokenBalances.ts
staylorwr: I think that by hand writing the `useTokenBalancesInterface` that info is all inferred.

src/views/Wallet/WalletStaking.vue
alexandreaco: 😅

PR #299, opened 2021-09-17
chore: Network switching improvements

alexandreaco:
Fixes the following stories on linear

- [chore: Show disclaimer if not on Mainnet network](https://linear.app/township/issue/RDX-245/feat-a-persistent-warning-bar-should-be-presented-when-users-are-on)
- [chore: Mainnet should appear before Stokenet](https://linear.app/township/issue/RDX-243/network-selection-mainnet-should-appear-above-stokenet)
- [chore: Explorer urls should be reactive to the active network ](https://linear.app/township/issue/RDX-241/switching-networks-should-update-the-url-used-for-explorer-links)
- [fix: overflow-x was overriding y axis and cropping tooltip](https://linear.app/township/issue/RDX-242/bug-user-created-token-balance-mouseover-is-partially-obscured)
- [chore: Expose invalid password error on login](https://linear.app/township/issue/RDX-244/bug-no-error-message-is-shown-when-an-incorrect-password-is-entered)


src/composables/useExplorerUrl.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface useExplorerUrlInterface {
+  readonly explorerUrlBase: ComputedRef<string>;
+  explorerUrlUnsub: () => void;
+}
+
+export default function useExplorerUrl (radix: RadixT): useExplorerUrlInterface {
+  const explorerUrlBase: Ref<string> = ref('')
+
+  const explorerUrlSub = radix.ledger.networkId().subscribe((n: Network) => {
```
staylorwr, line 12: This value should already be present on `useRadix`, I think I'd prefer for the explorer url to live there rather then extend a new subscription.

src/views/Home/index.vue
```diff
@@ -79,6 +81,7 @@ const Home = defineComponent({
 
     const authenticate = async (password: string) => {
       isAuthenticating.value = true
+      authenticatingError.value = false
```
staylorwr, line 84: 🤔 Is there a reason you're not pulling the `invalidPasswordError` ref from `useWallet` in the `HomePasswordComponnet`?  Whats the advantage of using the `rxRef` here when we have the error as part of the interface?  

As you mentioned in slack, I'm also ok with moving the logic down a layer to the `HomePasswordComponent` so we can get rid of the `rxRef` (I think this might be the last use of it?  We should get rid of it, we don't actually use the package for it's intended purpose)  That would also clean up this logic.

src/views/Wallet/WalletHistory.vue
```diff
@@ -110,6 +112,7 @@ const WalletHistory = defineComponent({
       verifyHardwareWalletAddress,
       activeAccount
     } = useWallet(radix, router)
+    const { explorerUrlBase, explorerUrlUnsub } = useExplorerUrl(radix)
```
staylorwr, line 115: Yeah seeing this again I think I'd prefer to have this computed property come directly from `useRadix`

src/views/Wallet/WalletOverview.vue
```diff
@@ -43,41 +43,12 @@
 
     <div class="bg-white text-rBlack py-4 px-8 flex-1">
       <div class="grid grid-cols-2 gap-4">
-        <div
+        <other-token-balance-list-item
```
staylorwr, line 46: 👌🏼

staylorwr (changes requested):
Styles are looking great!  Just a few nits about code style / composables

staylorwr (approved):

PR #304, opened 2021-09-20
fix: Staking transactions should reference nativeToken not selectedCurrency

alexandreaco:
This fixes a bug where the WalletConfirmTransactionModal would crash for both staking and unstaking


staylorwr (approved):
My favorite kind of PR 👌🏼

PR #306, opened 2021-09-20
fix: migration should represent full shape of wallet store before manipulating values

alexandreaco:


staylorwr (changes requested):
Starting from a local file store w/o a `wallet.json`:

* Remove the `wallet.json`
* Open app
* Screen shows a `Enter Passcode` page rather then UI for creating a new one.
* Entering a password throws an error
* 
Screen Shot 2021-09-20 at 1 22 00 PM


I think the trick will be to not run the migration if there is no store present @alexandreaco

alexandreaco:
@staylorwr I fixed the bug. I was setting the wallet to `{}` and not `""`. `{}` resolves to true in js so the app thought there was an active keystore. 

To use the migrations the way electron-store intended we should be writing migrations any time we add new values to the store. So we definitely want to run these on the initial step.

staylorwr (approved):

PR #312, opened 2021-09-21
fix: the cancel button should close tx conf modal

caleboller:
This PR fixes the cancel button on the transaction confirmation modal, correctly dismissing the modal.


saminakh (commented):
These changes LGTM, although I'm unable to test locally.

Vue is a weird place....trying to see _how_ the modal actually gets closed and chasing down logic in `useTransactions.ts` felt a bit challenging! But I think I understand a little better now.

saminakh (approved):

PR #313, opened 2021-09-21
chore: update styling of non-xrd balances

caleboller:
This PR updates non-XRD balances to have smaller type sizes, so they don't dominate the balances page.

Before
Screen Shot 2021-09-21 at 2 17 19 PM

After
Screen Shot 2021-09-21 at 2 11 21 PM


saminakh (approved):
Looks reasonable to me!

PR #317, opened 2021-09-22
fix: correctly present build errors on Transaction and Stake/Unstake screens

caleboller:
This PR sets up the the transaction and stake/unstake views to correctly watch the `transactionErrorMessage` value from the `useTransaction` composable to watch for any potential errors, and displays them appropriately.

It also resets the value of `transactionErrorMessage` to null in `useTransaction` when any new transaction is created, so that repeated failures are reported to views watching `transactionErrorMessage`.
Screen Shot 2021-09-21 at 9 36 32 PM


staylorwr (approved):
✨

PR #319, opened 2021-09-22
chore: Show default network ids even when offline

alexandreaco:
Screen Shot 2021-09-22 at 9 31 01 AM


caleboller (approved):
Thanks Alex! Custom Node URLs worked for me in this instance, and I was able to test. Confirmed this worked as expected both online and offline.

PR #330, opened 2021-09-22
feat: also disable change pin when no connection

caleboller:
This PR just disables the Change Pin setting tab when there's no connection present.


SK-Sam (approved):
👍

PR #332, opened 2021-09-23
feat: Expand network switch confirmation modal

staylorwr:
https://www.loom.com/share/dab2d555a6734436a3a3f07196133f5d


src/composables/useConnectableRadix.ts
```diff
@@ -26,17 +27,19 @@ export default function useConnectableRadix (): useConnectableRadixInterface {
...
-          isLoading.value = false
-          network.value = n
-        },
-        error: () => {
-          isConnected.value = false
-          isLoading.value = false
-        }
-      }))
+      subs.add(
+        tempRadix.ledger.networkId().pipe(timeout(5000)).subscribe({
```
alexandreaco, line 31: What does this timeout do?

staylorwr: https://rxjs.dev/api/operators/timeout

It throws an error if an event on the observable isn't created in a given amount of time.

alexandreaco (approved):
Pulled it down and this works for me 👍

PR #339, opened 2021-09-24
fix: show the selected currency in the conf modal

caleboller:
In the Great Refactor we introduced a regression where on the Amount in the transaction confirmation modal we always displayed the `nativeToken` symbol, where we should be showing the transaction's `selectedCurrency` symbol. This PR fixes that issue. It also ensures that we show the `nativeToken` symbol when staking/unstaking. 

Screen Shot 2021-09-24 at 8 32 01 AM

Also, there was a secondary issue on the create transaction form. The view was set up to watch the value of `nativeToken` and set it as the top (selected) item in the currency list every time the value changed. This meant that for every successful poll, XRD would bump to the top, replacing whatever the user had selected. This issue has also been patched.

### DIFF
--- a/src/views/Wallet/WalletConfirmTransactionModal.vue
+++ b/src/views/Wallet/WalletConfirmTransactionModal.vue
@@ -65,7 +65,8 @@
             <div class="w-26 text-right text-rGrayDark mr-6">{{ $t('transaction.amountLabel') }}</div>
             <div class="flex-1 flex flex-row items-center">
               <big-amount :amount="amount" class="mr-1" />
-              <span class="uppercase" v-if="nativeToken">{{ nativeToken.symbol }}</span>
+              <span class="uppercase" v-if="stakeInput && nativeToken">{{ nativeToken.symbol }}</span>
+              <span class="uppercase" v-else>{{ selectedCurrency.token.symbol }}</span>
             </div>
           </div>
 
@@ -182,7 +183,8 @@ const WalletConfirmTransactionModal = defineComponent({
       transactionFee,
       transactionState,
       transferInput,
-      transactionUnsub
+      transactionUnsub,
+      selectedCurrency
     } = useTransactions(radix, router, activeAccount.value, hardwareAccount.value, {
       ledgerSigningError: () => {
         hardwareAccountFailedToSign()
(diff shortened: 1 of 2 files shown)


alexandreaco (changes requested):
Screen Shot 2021-09-24 at 8 44 41 AM

This is going to require more to fix. That modal is shared between both the Transaction and Staking flow. We need to show the `nativeToken` when staking/unstaking and `selectedCurrency` when building a transfer

src/views/Wallet/WalletTransaction.vue
```diff
@@ -167,6 +167,7 @@ const WalletTransaction = defineComponent({
     const { t } = useI18n({ useScope: 'global' })
     const { nativeToken, nativeTokenUnsub } = useNativeToken(radix)
     const { tokenBalances, tokenBalanceFor, tokenBalancesUnsub } = useTokenBalances(radix)
+    let nativeTokenLoaded = false
```
alexandreaco, line 170: Can you please update this to use a ref so that it's truly reactive?
`const nativeTokenLoaded: Ref<boolean> = ref(false)`
and down below...
`nativeTokenLoaded.value = true`

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -65,7 +65,8 @@
             <div class="w-26 text-right text-rGrayDark mr-6">{{ $t('transaction.amountLabel') }}</div>
             <div class="flex-1 flex flex-row items-center">
               <big-amount :amount="amount" class="mr-1" />
-              <span class="uppercase" v-if="nativeToken">{{ nativeToken.symbol }}</span>
+              <span class="uppercase" v-if="stakeInput && nativeToken">{{ nativeToken.symbol }}</span>
+              <span class="uppercase" v-else>{{ selectedCurrency.token.symbol }}</span>
```
alexandreaco, line 69: Let's take what we learned about `nativeToken` being null and thowing an error when we display `selectedCurrency`
```
<span class="uppercase" v-if="stakeInput && nativeToken">{{ nativeToken.symbol }}</span>
<span class="uppercase" v-else-if="selectedCurrency">{{ selectedCurrency.token.symbol }}</span>
```

alexandreaco (approved):
thank you caleb!

PR #344, opened 2021-09-27
chore: Enable Latest Builds

staylorwr:
This PR enables the latest builds based on commits to main while preventing the build process from conflicting with the release cycle that could be running in parallel.

On release of 1.2.0, we discovered an issue where `release-please` and `release-latest` were racing to provide packaged builds that would be linked with the official release.

This PR removes that race condition by using a custom action to mark the `release-latest` builds as 'latest'.  This custom action does this by modifying the `package.json` file.

This enables:
* `vue.config.js` updates the `html title` of the electron app to include the 'LATEST' flag.
* `vue.config.js` overrides the default release behavior, setting it to null (there's a combination of bugs and poor documentation in `electron-builder` on the correct arguments to override publish feature) so `release-latest` will never target the tags and releases on github.
* 
Screen Shot 2021-09-27 at 1 44 01 PM

### DIFF
--- a/.github/actions/mark-latest/action.yml
+++ b/.github/actions/mark-latest/action.yml
@@ -1,0 +1,6 @@
+name: "mark-latest"
+description: "Mark the release as latest in package.json"
+
+runs:
+  using: "node12"
+  main: "index.js"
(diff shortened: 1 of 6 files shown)


alexandreaco (approved):
This is a great solution, thank you :)

PR #349, opened 2021-10-04
feat: Add error modals for generic and txn errors

alexandreaco:
This PR... 
- adds a reusable modal component that can be expanded upon for more standard errors across the app
- adds a new composable called `useErrors` that manages the list of errors the app has caught
- Displays an error to the user when building a transaction fails
- Displays an error to the user when finalizing a transaction fails

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -3,8 +3,18 @@
     <div class="bg-rBlue w-1/2 h-tallest origin-center transform rotate-45 absolute z-0 top-0 left-0 -translate-x-1/2 -translate-y-1/2"></div>
     <div class="relative z-10 min-h-screen">
       <template v-if="latestError">
+        <ErrorModalTransactionBuilding
+          v-if="latestError.type === 'TRANSACTION-BUILDING'"
+          :error="latestError.error"
+          :errorsCount="errorsCount"
+        />
+        <ErrorModalTransactionConfirming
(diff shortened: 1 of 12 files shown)


caleboller:
@alexandreaco How do you feel about reverting the general error sink modals? The error modal behavior isn't appealing, and I don't think it's terribly useful in the situation. When QAing TX error modals I'd regularly be hit with the general errors before/during/after seeing the error related to the work I'm doing, causing even more confusion.

If an error occurs in some background fetch/task, it should not interrupt the user. The user should only see error modals if they're prompted by an explicit user action, like a button click.

alexandreaco:
@caleboller I agree with not showing the error sink errors as modals. It does really take over the user experience. I updated the PR!

alexandreaco:
@caleboller I removed the inline validation error on the transaction screen. Now, when you naively attempt to send more tokens than you have we show a modal with the appropriate error message instead of the inline red text. I think this is a nice UX.

For num. 2, this is a more complex ask. Right now we are actually canceling that transfer request when it returns an error. Leaving the user on the modal would mean that we are not cancelling the request but instead leaving it open. I'd prefer not to make a change like that right before we go on retreat as it may create more unexpected behavior.

caleboller (approved):

PR #350, opened 2021-10-05
chore: Differentiate in the UI between loading data and the user having no data

alexandreaco:
This PR updates three different views

1. The History view should show a loading indicator while the data is loading and only show "No transactions" when the sub resolves with no transactions
2. The Transaction view should show a loading indicator while the data is loading and only show "No tokens" when the sub resolves with no tokens
3. The Staking view should show a loading indicator in the sidebar while the data is loading

https://linear.app/township/issue/RDX-279/better-error-messages-for-slow-loading-history-send-and-stake-views

### DIFF
--- a/src/composables/useStaking.ts
+++ b/src/composables/useStaking.ts
@@ -1,23 +1,35 @@
-import { ref, Ref } from 'vue'
+import { ref, computed, ComputedRef, Ref } from 'vue'
 import { RadixT, StakePositions, UnstakePositions } from '@radixdlt/application'
 
 interface useStakingInterface {
   readonly activeStakes: Ref<StakePositions | null>;
   readonly activeUnstakes: Ref<UnstakePositions | null>;
+  readonly loadingAllStaking: ComputedRef<boolean>;
   stakingUnsub: () => void;
 }
 
 export default function useStaking (radix: RadixT): useStakingInterface {
   const activeStakes: Ref<StakePositions | null> = ref(null)
   const activeUnstakes: Ref<UnstakePositions | null> = ref(null)
+  const loadingStakes: Ref<boolean> = ref(true)
+  const loadingUnstakes: Ref<boolean> = ref(true)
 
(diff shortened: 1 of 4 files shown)


SK-Sam:
💯 
Screen Shot 2021-10-05 at 12 25 52 PM

src/composables/useStaking.ts
```diff
@@ -1,23 +1,35 @@
...
-  const activeStakesSub = radix.stakingPositions.subscribe((stakes: StakePositions) => { activeStakes.value = stakes })
-  const activeUnstakesSub = radix.unstakingPositions.subscribe((unstakes: UnstakePositions) => { activeUnstakes.value = unstakes })
+  const activeStakesSub = radix.stakingPositions.subscribe((stakes: StakePositions) => {
+    activeStakes.value = stakes
+    loadingStakes.value = false
+  })
+  const activeUnstakesSub = radix.unstakingPositions.subscribe((unstakes: UnstakePositions) => {
+    activeUnstakes.value = unstakes
+    loadingUnstakes.value = false
+  })
```
SK-Sam, line 24: a bit of a naiive question as the first time I used a Ref was to refer to an SVG, but why do we initialize `loadingStakes` and `loadingUnstakes` with a ref true? is this due to what the SDK demands?

SK-Sam (approved):
This is so cool! I tested this on both Mainnet w/ an empty wallet, and Stokenet w/ loadable content. I did feel the loading wheel is such a big QOL improvement 😄 cool stuff!!

src/composables/useStaking.ts
alexandreaco: Using a `ref` makes this a reactive value. This means that Vue can watch for changes under the hood and correctly pass them to components who call `useStaking()`

SK-Sam: cool, thank you 😄 TIL!!!!

PR #352, opened 2021-10-05
chore: change tokenbalance identifier name to token name

SK-Sam:
This PR changes what's being populated in the Title of each otherTokenBalance. It used to use the token symbol and is now using the token's name.

Before:
Screen Shot 2021-10-05 at 1 43 50 PM

After:
Screen Shot 2021-10-05 at 1 44 03 PM


alexandreaco (approved):
🔥 that's it!

PR #357, opened 2021-10-06
feat: show copy when verifying non-active hw acct

caleboller:
When a user attempts to copy the non-active hardware account account number, we now show appropriate copy instead of the verify process. See video for details.

https://www.loom.com/share/bf3c3800d1b34d019daf8dc109b79d1d


src/views/Wallet/WalletLedgerVerifyAddressModal.vue
```diff
@@ -2,27 +2,42 @@
...
+      <div v-if="hardwareAddress !== activeAddress.toString()">
+        <div class="py-2">
+          Whenever copying a hardware wallet address, it is strongly recommended to verify it on the hardware device. To copy and verify this address, please switch to the hardware address first by selecting it in the account picker.
+        </div>
+        <div class="flex items-center">
+          <div
+            class="text-rGrayDark py-3 text-base py-12 mx-auto cursor-pointer"
+            @click="hideLedgerVerify()"
+          >
+            Dismiss
```
saminakh, line 14: A small nit here, but I personally feel like the dismiss should either: 1) have less spacing around it as long as it's just a clickable div, or 2) be converted into a button so there isn't such an excess of space!

saminakh (approved):
The code LGTM! i left a few style nits, but nothing super critical!

PR #371, opened 2021-10-26
fix: prevent crash on sleep and hide max button

caleboller:
This PR does two small things:

- Currently, if a user clicks the `X` button on OS X only, and then put their computer to sleep they'll return a crash report. This fix prevents that crash from occurring. We were attempting to reset state in a window that did not exist after being closed, we prevent that from occurring.
- It also disables the maximize button which was causing confusion among users.


caleboller:
@alexandreaco @staylorwr  I QA'd this and it looks good to me! Mergin'.

PR #374, opened 2021-11-04
chore: show warning when txn brings total XRD < 10

SK-Sam:
[Demo for Sending/Staking](https://www.loom.com/share/a91f49fb5c0d4e178a5c51edd8cc078d)
[Demo for Unstaking & the lack of this warning message](https://www.loom.com/share/b2823102ee4f4b9e94f16625a5180eaf)

This PR will throw a warning to the User to let them know upon this transaction, their total XRD balance will be < 10 XRD. This message applies to Staking and Sending Tokens, but not Unstaking.

Warning message should apply to case where User sends NON XRD token as well, since there is still a `Transaction Fee` involved.
- Sending/Staking XRD Calculation:
   -  Total XRD - XRD Amount Input(to stake or to send) - Transaction Fee > 10?
- Sending NON XRD Calculation:
   - Total XRD - Transaction Fee > 10?
- Unstaking should never trigger this warning, even if...User has 2 XRD and is unstaking 1 XRD. Even if this is < 10 XRD we don't want to warn them in the case of unstaking.

Since we're hitting the API to get User's Total XRD Balance in the WalletConfirmTransactionModal, would it be helpful to display the User's total XRD post-transaction as well? Another row that says...`Remaining XRD After Transaction: ` `1,0004.068 XRD`? something like this? Would like your thoughts on this @alexandreaco @saminakh Example:

Screen Shot 2021-11-04 at 7 33 18 PM

### DIFF
--- a/src/views/Wallet/WalletConfirmTransactionModal.vue
+++ b/src/views/Wallet/WalletConfirmTransactionModal.vue
@@ -97,7 +97,10 @@
           </div>
         </div>
 
-        <div class="pr-6" v-if="xrdRemainderBalanceBelowTen">
+        <div
+          v-if="xrdRemainderBalanceBelowTen"
+          class="pr-6"
+        >
           <span class="text-rRed text-md"><b>{{ $t('transaction.warningTitle') }}:</b> {{ $t('transaction.lessThanTenXRDBalanceWarning') }}</span>
         </div>
 
@@ -193,10 +196,10 @@ const WalletConfirmTransactionModal = defineComponent({
 
     const balanceSub = updateObservable.pipe(
       switchMap(() => radix.activeAccount),
-      mergeMap((account) => forkJoin([
+      mergeMap((account) =>
         radix.ledger.tokenBalancesForAddress(account.address)
-      ]))
-    ).subscribe(([balances]) => {
+      )
+    ).subscribe((balances) => {
       tokenBalances.value = balances
       loading.value = false
     })
--- a/src/views/Wallet/WalletConfirmTransactionModal.vue
+++ b/src/views/Wallet/WalletConfirmTransactionModal.vue
@@ -97,6 +97,10 @@
           </div>
         </div>
 
+        <div class="pr-6" v-if="remainderBalanceAboveTen">
+          <span class="text-rRed text-md"><b>Warning:</b> This transaction will leave you with less than 10 XRD. Remember that you must have XRD to pay fees to unstake and perform other transactions.</span>
+        </div>
+
         <div class="flex items-start justify-between w-full mt-1">
           <div class="flex items-start justify-start mt-1 flex-1">
             <div class="font-light text-rGrayDark bg-rGrayLight border border-rGray py-2 pl-4 pr-5 rounded text-base w-30">
@@ -132,14 +136,16 @@
 </template>
 
 <script lang="ts">
-import { AmountOrUnsafeInput } from '@radixdlt/application'
-import { defineComponent, ref, onMounted, onUnmounted, computed, ComputedRef } from 'vue'
+import { Amount, AmountT, AmountOrUnsafeInput, SimpleTokenBalance, SimpleTokenBalances, Token, StakePositions } from '@radixdlt/application'
+import { defineComponent, ref, Ref, onMounted, onUnmounted, computed, ComputedRef } from 'vue'
 import { useForm } from 'vee-validate'
+import { merge, interval, Subscription, forkJoin } from 'rxjs'
+import { switchMap, mergeMap } from 'rxjs/operators'
 import BigAmount from '@/components/BigAmount.vue'
 import PinInput from '@/components/PinInput.vue'
 import ButtonSubmit from '@/components/ButtonSubmit.vue'
 import { validatePin } from '@/actions/vue/create-wallet'
-import { useRouter } from 'vue-router'
+import { useRouter, onBeforeRouteLeave } from 'vue-router'
 import { useNativeToken, useHomeModal, useTransactions, useWallet } from '@/composables'
 import { useI18n } from 'vue-i18n'
 
@@ -163,7 +169,6 @@ const WalletConfirmTransactionModal = defineComponent({
     const isValidPin = ref(false)
     const pinAttempts = ref(0)
     const { t } = useI18n({ useScope: 'global' })
-
     const {
       activeAddress,
       activeAccount,
@@ -173,6 +178,35 @@ const WalletConfirmTransactionModal = defineComponent({
       reset
     } = useWallet(router)
     const { nativeToken, nativeTokenUnsub } = useNativeToken(radix)
+
+    const updateObservable = merge(
+      radix.activeAccount,
+      interval(15000)
+    )
+
+    const subs = new Subscription()
+    const loading = ref(true)
+
+    subs.add(radix.activeAddress.subscribe(() => {
+      loading.value = true
+    }))
+
+    const balanceSub = updateObservable.pipe(
+      switchMap(() => radix.activeAccount),
+      mergeMap((account) => forkJoin([
+        radix.ledger.tokenBalancesForAddress(account.address),
+        radix.ledger.stakesForAddress(account.address),
+        radix.ledger.unstakesForAddress(account.address)
+      ]))
+    ).subscribe(([balances, stakes, unstakes]) => {
+      tokenBalances.value = balances
+      loading.value = false
+    })
+    subs.add(balanceSub)
+
+    onBeforeRouteLeave(() => {
+      subs.unsubscribe()
+    })
 
     const {
       activeMessageInTransaction,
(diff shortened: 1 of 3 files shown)


src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -174,6 +179,35 @@ const WalletConfirmTransactionModal = defineComponent({
...
+
+    subs.add(radix.activeAddress.subscribe(() => {
+      loading.value = true
+    }))
+
+    const balanceSub = updateObservable.pipe(
+      switchMap(() => radix.activeAccount),
+      mergeMap((account) => forkJoin([
+        radix.ledger.tokenBalancesForAddress(account.address),
+        radix.ledger.stakesForAddress(account.address),
```
alexandreaco, line 198: We don't need stakesForAddress and unstakesForAddress here so we can simplify this subscription.

```js
switchMap(() => radix.activeAccount),
mergeMap((account) => radix.ledger.tokenBalancesForAddress(account.address))
```

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -174,6 +179,35 @@ const WalletConfirmTransactionModal = defineComponent({
...
+        radix.ledger.unstakesForAddress(account.address)
+      ]))
+    ).subscribe(([balances, stakes, unstakes]) => {
+      tokenBalances.value = balances
+      loading.value = false
+    })
+    subs.add(balanceSub)
+
+    onBeforeRouteLeave(() => {
+      subs.unsubscribe()
```
alexandreaco, line 208: Good thinking, we always want to unsubscribe. However, this view isn't a route view, it's a nested component. So here, let's use `onUnmounted()` instead of `onBeforeRouteLeave()`

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -297,9 +347,24 @@ const WalletConfirmTransactionModal = defineComponent({
...
+      totalXRD,
       values,
       selectedCurrency
     }
+  },
+  computed: {
+    // Do not check for balance above 10XRD if unstaking. Only run validation if staking/sending. Two possible calculations
+    // can occur. Either (Total XRD - Transaction Fee) OR (Total XRD - XRD Amount Input - Transaction Fee). Calculation is
+    // based on whether or not User chose to send/stake XRD or another token.
+    xrdRemainderBalanceAboveTen (): boolean {
```
alexandreaco, line 359: Do you mean "below" ten? :) 

Also, please move this to a computedRef within setup() to keep with the composition api spec.

alexandreaco (changes requested):
Thanks for the inline docs and thorough PR comments! I left some changes to clean this up. Great first pass :)

src/views/Wallet/WalletConfirmTransactionModal.vue
SK-Sam: I definitely meant `below` and my results are making much more sense now w/ the name change 😸

SK-Sam:
🙇 🙇 changed up to your specs, thanks for the kind review @alexandreaco ! Didn't realize `ComputedRef` was part of the composition api.

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -174,6 +179,33 @@ const WalletConfirmTransactionModal = defineComponent({
...
+    const subs = new Subscription()
+    const loading = ref(true)
+
+    subs.add(radix.activeAddress.subscribe(() => {
+      loading.value = true
+    }))
+
+    const balanceSub = updateObservable.pipe(
+      switchMap(() => radix.activeAccount),
+      mergeMap((account) => forkJoin([
```
alexandreaco, line 196: You don't need the forkJoin here because we only have one sub now :)

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -174,6 +179,33 @@ const WalletConfirmTransactionModal = defineComponent({
...
+        radix.ledger.tokenBalancesForAddress(account.address)
+      ]))
+    ).subscribe(([balances]) => {
```
alexandreaco, line 199: When you remove forkJoin you'll also take this out of the array deconstructor

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -97,6 +97,10 @@
           </div>
         </div>
 
+        <div class="pr-6" v-if="xrdRemainderBalanceBelowTen">
```
alexandreaco, line 100: I'm going to leave the same style nit here that I put on samina's PR so we can be more consistent moving forward. 

When you have a DOM element with a vue attr, you want to break up that div onto separate lines. v-if and other templating attrs should always come first

```jsx
<div
  v-if="xrdRemainderBalanceBelowTen"
  class="pr-6"
>
  <span class="text-rRed text-md">
    <b>{{ $t('transaction.warningTitle') }}:</b> {{ $t('transaction.lessThanTenXRDBalanceWarning')}}
  </span>
</div>
```

SK-Sam: good to know 👍

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -272,6 +307,36 @@ const WalletConfirmTransactionModal = defineComponent({
       isValidPin.value = false
     }
 
+    const tokenBalances: Ref<SimpleTokenBalances | null> = ref(null)
+    const tokenBalanceFor = (token: Token) => {
+      if (!tokenBalances.value) return null
+      return tokenBalances.value.tokenBalances.find((tb: SimpleTokenBalance) => tb.tokenIdentifier.equals(token.rri)) || null
+    }
+
+    const zero = Amount.fromUnsafe(0)._unsafeUnwrap()
```
saminakh, line 316: 👍 this is handy!

saminakh (commented):
The warning looks good and the logic looks sound! I did notice one case that may have slipped through the cracks: when a user has less than 10 XRD to begin with, and makes a transaction that keeps their balance below 10 XRD. I think no warning is supposed to show in this case, b/c at that point _every_ transaction will cause them to be below 10 because they were below 10 to begin with!
Screen Shot 2021-11-05 at 2 52 37 PM

SK-Sam:
> The warning looks good and the logic looks sound! I did notice one case that may have slipped through the cracks: when a user has less than 10 XRD to begin with, and makes a transaction that keeps their balance below 10 XRD. I think no warning is supposed to show in this case, b/c at that point _every_ transaction will cause them to be below 10 because they were below 10 to begin with! Screen Shot 2021-11-05 at 2 52 37 PM

this may be true, i probably wouldn't want someone telling me every transaction how close to 0 i am 😢

SK-Sam:
> The warning looks good and the logic looks sound! I did notice one case that may have slipped through the cracks: when a user has less than 10 XRD to begin with, and makes a transaction that keeps their balance below 10 XRD. I think no warning is supposed to show in this case, b/c at that point _every_ transaction will cause them to be below 10 because they were below 10 to begin with! Screen Shot 2021-11-05 at 2 52 37 PM

Well hmm, thought about it over the weekend. if the main point of the warning is to warn the user that there's an underlying fee to transactions, should this warning be thrown even when under 10 XRD? 🤔 💭 it mayyy be safe to assume the user is new to the Radix platform?

alexandreaco:
Hey @SK-Sam, I checked the original Linear story and Matt would like the warning to **not** appear when a user has less than 10 xrd to begin with
> Let's just show a bright red warning in the confirmation modal if a user is ever about the submit a TX that would take their XRD balance from > 10 XRD to < 10 XRD.

https://linear.app/township/issue/RDX-252/add-warning-message-when-about-to-send-all-or-nearly-all-xrd

alexandreaco:
@saminakh Can you please re-review this? I made the last update to not show the warning when users have < 10 xrd to begin with

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -328,8 +328,13 @@ const WalletConfirmTransactionModal = defineComponent({
     // based on whether or not User chose to send/stake XRD or another token.
     // Default value is false because we don't want to blindly throw this warning outside of said cases.
     const xrdRemainderBalanceBelowTen: ComputedRef<boolean> = computed(() => {
+      // Check transaction fee doesn't bring total < 10 xrd for non-exd transactioins
       if (selectedCurrency.value && selectedCurrency.value.token.symbol !== 'xrd' && toLabel.value !== 'Unstake from') {
         return Number(totalXRD.value) - Number(transactionFee.value) < Math.pow(10, 19)
+      // Don't show error if starting total is < 10 xrd
+      } else if (totalXRD.value && Number(totalXRD.value) < Math.pow(10, 19)) {
```
saminakh, line 335: My only thought here is, should this be the top case? The case itself looks correct though!

saminakh (approved):
Left one Q, but leaving an approval to speed up the merging process in case my comment isn't applicable! The case itself LGTM!

src/views/Wallet/WalletConfirmTransactionModal.vue
alexandreaco: You're right, thank you! Updated :)

alexandreaco (approved):

PR #377, opened 2021-11-15
chore: add indication to current stakes list

SK-Sam:
[Demo](https://www.loom.com/share/3bb6bef1a1fa4bdab4a01a1a529034a9)

This PR gets the Top 100 Validators, and checks against User's current stake positions. If any positions are outside of the Top 100 Validators, we will indicate by showing a `no emissions` text on validator.

Demo video mocks a Top 28 Validators list, and I want to show that by changing the value to 27, the 28th Validator will show the `no emissions` text.

### DIFF
--- a/src/views/Wallet/WalletStaking.vue
+++ b/src/views/Wallet/WalletStaking.vue
@@ -88,7 +88,7 @@
           </svg>
         </a>
       </div>
-      <div v-if="loadingAllStaking || !nativeToken" class="p-4 flex items-center justify-center">
+      <div v-if="loadingAnyStaking || !nativeToken" class="p-4 flex items-center justify-center">
         <loading-icon class="text-rGrayDark" />
       </div>
       <template v-else>
@@ -99,6 +99,7 @@
           :nativeToken="nativeToken"
           :explorerUrlBase="explorerUrlBase"
           :notTopOneHundred="!validatorsTopOneHundred.includes(position)"
+          :preloadedValidator="maybeGetPreloadedValidator(position)"
           @addToValidator="handleAddToValidator"
           @reduceFromValidator="handleReduceFromValidator"
         >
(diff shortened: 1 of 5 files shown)


saminakh (commented):
I think the code overall looks good! I think it would be a nice-to-have if there wasn't a delay between the page rendering and the app fetching the top 100 list/determining the 'no emissions' value. When i click on the staking/unstaking view, it indicates the few validators i have stakes w/ are 'no emissions', but then after a few moments the indicator disappears. Would swapping the default help with this? Also, would it make sense to subscribe to the top 100 list earlier in the flow? (Maybe when the app boots up, rather than when you first open the staking/unstaking page?) I'm not sure if that's doable, perhaps @alexandreaco or @staylorwr would have suggestions here.

src/views/Wallet/WalletStaking.vue
```diff
@@ -161,6 +163,21 @@ const WalletStaking = defineComponent({
...
+    radix.ledger
+      .validators({ size: 100 })
+      .subscribe((validatorsRes: Validators) => {
+        validators.value = validatorsRes
+      })
+      .add(subs)
+
+    onUnmounted(() => {
+      subs.unsubscribe()
+    })
```
staylorwr, line 179: @SK-Sam this looks like the right logic but it seems we have a little cleaning up to do here.  Can you move this validator logic into the `useStaking` composable? 

We should also only have a single call to `useStaking` in the setup of this function, as we're currently calling it twice, and generating two sets of `radix.stakingPositions` and `radix.unstakingPositions` subscriptions but only unsubscribing from a single one.   (line 158 and line 183).  

Once you have all of the subscription logic in `useStaking`, could you also take a look at moving the reference to the global list of top 100 validators outside of the `useStaking` function definition?   `useWallet` and `useTransactions` have some examples of defining the reference outside of the function definition and then passing those const references.  (This allows for the reference to exist outside of the lifecycle of the composable and the component that called the composable, making them effectively 'global' values that can be referenced multiple times without losing context, like the main `radix` instance from `useWallet`)

Otherwise this looks great, sorry that we still have some bad examples of non-composable logic floating around.  Thank you @SK-Sam!

staylorwr (changes requested):

src/views/Wallet/WalletStaking.vue
alexandreaco: I went in and cleaned this up. Can you give it another review?

staylorwr (approved):

PR #378, opened 2021-11-16
fix: Cleanup Transaction Subscriptions when a transaction is cancelled

staylorwr:


caleboller (approved):
@staylorwr From a QA perspective, this behaves well. It resolves the issue described in RDX-296, and doesn't introduce any regressions so far as I could tell. I QA'd both Send Tokens and Stake/Unstake.

src/composables/useTransactions.ts
```diff
@@ -190,8 +190,10 @@ export default function useTransactions (radix: RadixT, router: Router, activeAc
   const setActiveTransactionForm = (val: string) => {
     activeTransactionForm.value = val
   }
-
+  let activeTransactionSubs = new Subscription()
   const confirmAndExecuteTransaction = (transactionTracking: TransactionTracking) => {
+    activeTransactionSubs.unsubscribe()
+    activeTransactionSubs = new Subscription()
```
saminakh, line 196: I don't think I'm understanding this code, could you explain it to me? It looks like `subs` and `activeTransactionSubs` are subscribed to all the same things at the same times, but you're only unsubscribing `activeTransactionSubs` after cancellation? Why don't `subs` need to unsubscribe as well? (Sorry, subscriptions are very new to me!)

saminakh (commented):
I don't think I can leave helpful feedback on this one, sorry! I don't know enough about subscriptions to help!

src/composables/useTransactions.ts
```diff
@@ -245,13 +249,16 @@ export default function useTransactions (radix: RadixT, router: Router, activeAc
...
 
     // Store draft transaction actions
-    subs.add(transactionTracking.events
+    const draftTransactionSub = transactionTracking.events
       .pipe(filter((trackingEvent: TransactionStateUpdate) => trackingEvent.eventUpdateType === 'INITIATED'))
       .subscribe((res: TransactionStateUpdate) => {
         draftTransaction.value = (res as TransactionStateSuccess).transactionState as TransactionIntent
-      }))
+      })
+    subs.add(draftTransactionSub)
```
alexandreaco, line 260: 👍 good catch. this syntax has gotten us before

alexandreaco (approved):
Code changes make sense to me. It seems like we're cleaning up all of the subs now 👍

src/composables/useTransactions.ts
staylorwr: 👋🏼 Thanks for taking the time to review the code @saminakh, let me see how well I can explain this bug:

The `WalletConfirmTransaction` modal component runs the `setup` function _every time_ the component is rendered.  In the case of this flow, that's every time a transaction form is submitted either on the `WalletTransaction` or `WalletStaking` views.  

The problem here lies in what that `setup` function does.  Every time `useTransactions` is called, we build subscriptions to relevant events and life cycles of a transaction.  The bug occurs _when we don't cancel those subscriptions_.  

Because the modal is hidden on 'cancel' those error watchers or state change subs keep sitting in the background then race to run logic when an event comes through (these rxjs things have funny, specific names that I still don't fully understand).  Sometimes _that logic is waiting to send commands to the ledger device_, or waiting to throw errors.  

This PR adds a way for the confirmation modal to unsubscribe from those business logic subscriptions before unmounting.  That way, a modal is only maintaining a single set of subscriptions for a single set of ledger events.  There are probably a few ways this could be improved, including moving this logic up to a single set of initiating logic on the Wallet Index logic and passing the events, state, and methods down in props and events.  But given the scope of the work required, and the complexity of these flows that should change a bit with forthcoming SDK changes, I landed on this simpler solution.

PR #380, opened 2021-11-22
fix: custom node URL inputs should be trimmed

caleboller:
When a user submits a custom node URL, we should first trim their entry before comparing for uniqueness or adding the URL.

[See a demo](https://www.loom.com/share/9322f5c00a21426bb0f4b02172ff9a09)


saminakh (approved):
Code LGTM, tried it out locally too 👍

PR #381, opened 2021-11-22
fix: add conditional margin to avoid tooltip from being clipped by parent divs

SK-Sam:
[Demo](https://www.loom.com/share/ff8a4d37fb814e47a857be49df8033d7)

This PR addresses the issue of when tooltip amounts of 3rd party balances are clipped by its container. Here is an example.

Before:
Screen Shot 2021-11-22 at 11 24 57 AM

After: 
Screen Shot 2021-11-22 at 11 23 16 AM

 
As you can see here, the first digit is clipped. This occurs due to tooltips being based on the center of the display number. With the display number being a low amount of digits, the tooltip begins farther on the left as opposed to other longer digit numbers.

To handle this, we are offsetting the margin for numbers with digits lower than 4. In the demo, we are using an extreme case of a 1 digit display balance number, and a 17 digit tooltip balance. We don't want this offsetting to occur on long digit display balances either, so we are using a conditional to offset specifically low digit display balances.

Other approaches attempted:
- Z index doesn't affect tooltip as the overflow of the outermost div doesn't allow z index to go over.
- Changing outermost div's overflow style to `overflow-visible` allows for tooltip to never be clipped, but we lose the ability to scroll horizontally.

### DIFF
--- a/src/components/BigAmount.vue
+++ b/src/components/BigAmount.vue
@@ -1,7 +1,10 @@
 <template>
   <span class="relative inline-flex flex-col items-center group cursor-pointer" @click.stop="copyText">
     <span>{{numberForDisplay}}</span>
-    <div class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 rounded-sm shadow border border-solid border-rGrayLight">
+    <div v-if="numberLessThanFourDigits" class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 ml-32 rounded-sm shadow border border-solid border-rGrayLight">
+      {{fullNumber}}
+    </div>
+    <div v-else class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 rounded-sm shadow border border-solid border-rGrayLight">
       {{fullNumber}}
     </div>
   </span>
@@ -120,6 +123,14 @@ const BigAmount = defineComponent({
 
     fullNumber (): string {
       return asBigNumber(this.amount, true)
+    },
+
+    numberLessThanFourDigits (): boolean {
+      if (asBigNumber(this.amount, false).toString().length < 4) {
+        return true
+      } else {
+        return false
+      }
     }
   },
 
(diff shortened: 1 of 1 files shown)


lkoller (commented):
this looks good to me 👍

src/components/BigAmount.vue
```diff
@@ -120,6 +123,14 @@ const BigAmount = defineComponent({
 
     fullNumber (): string {
       return asBigNumber(this.amount, true)
+    },
+
+    numberLessThanFourDigits (): boolean {
+      if (asBigNumber(this.amount, false).toString().length < 4) {
```
caleboller, line 129: @SK-Sam Couldn't this be as simple as the following?
 `return asBigNumber(this.amount, false).toString().length < 4`

src/components/BigAmount.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <span class="relative inline-flex flex-col items-center group cursor-pointer" @click.stop="copyText">
     <span>{{numberForDisplay}}</span>
-    <div class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 rounded-sm shadow border border-solid border-rGrayLight">
+    <div v-if="numberLessThanFourDigits" class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 ml-32 rounded-sm shadow border border-solid border-rGrayLight">
```
caleboller, line 4: I think you could make this a good deal DRYer using Vue's `:class` binding, which allows you to have a bit more logic in your class definition. I _think_ you could leave the original class declaration as is, and then addition an additional `:class` that was something like:

`:class=numberLessThanFourDigits && 'ml-32'"` 

You can read more about `:class` [here](https://vuejs.org/v2/guide/class-and-style.html) and [here](https://michaelnthiessen.com/conditional-class-binding-vue).

caleboller (changes requested):

src/components/BigAmount.vue
saminakh: I think this is good feedback! Making a note for myself too, I'm sure this will come in handy at some point!

saminakh (commented):
Fix LGTM! I think caleb's suggestions are worth implementing, and then if you want to @ me for an approval I can do it!

src/components/BigAmount.vue
SK-Sam: thanks @caleboller , i actually see a few examples in our codebase now that I look at it 😄

src/components/BigAmount.vue
SK-Sam: why...yes it can! 😶‍🌫️

caleboller (approved):

PR #382, opened 2021-11-22
chore: disable Escape event listener earlier in lifecycle

SK-Sam:
This PR prevents User from closing the Transaction modal by hitting Escape. It gives a false sense that the tx has been cancelled when in fact it hasn't. For hardware wallet, the tx can still be cancelled via declining on said hardware wallet. On Olympia, the tx can't be cancelled after Confirm button has been hit(this isn't new behavior).

I've created a function which removes an event listener upon hitting the Confirm button. Using this lifecycle, should I remove the `windows.removeEventListener()` from `onUnmounted` hook, since it should be disabled by then?

### DIFF
--- a/src/views/Wallet/WalletConfirmTransactionModal.vue
+++ b/src/views/Wallet/WalletConfirmTransactionModal.vue
@@ -121,7 +121,7 @@
               {{ $t('transaction.cancelButton') }}
             </div>
 
-            <ButtonSubmit class="w-44 py-3" :disabled="disableSubmit" :small="true" >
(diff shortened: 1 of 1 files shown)


SK-Sam:
I'm not seeing a Cancel button render for Part 1 of this story(checkbox 1) re: Hardware ledgers. I've gone through the notions of

1: Creating a TX by sending token

2: Confirming PIN and hitting Confirm

3: While in the loading/Submitting TX, I still don't see any cancel buttons

4: While signing on my ledger, i still don't see any Cancel buttons render.

Let me know if i should be seeing a Cancel button render on Olympia

@caleboller @saminakh

caleboller (approved):
Great solution @SK-Sam! Works as expected, very clever.

PR #386, opened 2021-11-23
fix: don't clip token mouseovers

caleboller:
This PR better handles token mouseovers. The previous fix from @SK-Sam was clever, but didn't work in some edge cases where the presented token value was very short, like `0.106`, but the mouseover value was the full 18 digits. This solution allows the token mouseover to just extend outside the other token cell.
Screen Shot 2021-11-23 at 7 57 32 AM

### DIFF
--- a/src/components/OtherTokenBalanceListItem.vue
+++ b/src/components/OtherTokenBalanceListItem.vue
@@ -15,7 +15,7 @@
         </div>
       </div>
     </div>
-    <div class="overflow-x-scroll relative">
+    <div class="overflow-x-visible relative">
       <div class="flex flex-row absolute">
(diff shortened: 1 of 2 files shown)


src/components/OtherTokenBalanceListItem.vue
```diff
@@ -15,7 +15,7 @@
         </div>
       </div>
     </div>
-    <div class="overflow-x-scroll relative">
+    <div class="overflow-x-visible relative">
```
SK-Sam, line 18: approving due to short deadline, but I do see a style change in the token boxes by a bit (spacing between token name & token title

SK-Sam (approved):

PR #388, opened 2021-11-23
Fix: update pin copy

caleboller:
We had abused the `confirm` key that was used elsewhere when building the network selection page. I've made a unique key for the network confirmation button, and updated the reference.


SK-Sam (approved):
before: 
Screen Shot 2021-11-23 at 11 16 18 AM

after: 
Screen Shot 2021-11-23 at 11 17 04 AM 

Looks the same, great! LGTM 👍

PR #394, opened 2021-11-30
chore: remember last selected account

saminakh:
this PR:
- stores the address of the latest account selected on account switch
- switches accounts when the latest account saved is loaded and doesn't match the currently active account
https://www.loom.com/share/60ad0c46123f4c0297b53baf9847d781

### DIFF
--- a/src/actions/electron/data-store.ts
+++ b/src/actions/electron/data-store.ts
@@ -20,6 +20,10 @@ export const getLatestAccountAddress = (): string => {
   return store.get('latestAddress', '') as string
 }
 
+export const saveLatestAccountAddress = (event: IpcMainInvokeEvent, address: string) => {
+  return store.set('latestAddress', address)
+}
+
 export const getAccountNames = (): AccountName[] => {
   const accounts = store.get('account', {}) as { [key: string]: string; }
   const asAccounts: AccountName[] = []
--- a/src/actions/electron/data-store.ts
+++ b/src/actions/electron/data-store.ts
@@ -16,6 +16,10 @@ export const saveAccountName = (event: IpcMainInvokeEvent, data: string) => {
   return store.set(`account.${accountAddress}`, prettyName)
 }
 
+export const getLatestAccountAddress = (): string => {
+  return store.get('latestAddress', '') as string
+}
+
 export const getAccountNames = (): AccountName[] => {
   const accounts = store.get('account', {}) as { [key: string]: string; }
   const asAccounts: AccountName[] = []
--- a/src/actions/electron/data-store.ts
+++ b/src/actions/electron/data-store.ts
@@ -17,12 +17,12 @@ export const saveAccountName = (event: IpcMainInvokeEvent, data: string) => {
 }
 
 export const getLatestAccountAddress = (event: IpcMainInvokeEvent, network: string): string => {
-  return store.get(`latestAddress.${network}`, '') as string
+  return store.get(`wallets.${network}.latestAddress`, '') as string
 }
 
 export const saveLatestAccountAddress = (event: IpcMainInvokeEvent, data: string) => {
   const { address, network } = JSON.parse(data)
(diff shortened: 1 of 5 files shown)


alexandreaco (changes requested):
Great first pass. I can see that it remembers my last selected account. There are two things that we are still missing

1. We should store the last saved account per network the same way we are storing `derivedAccountsIndex`. That way I can have a latest MAINNET and STOKENET account
2. Don't forget to add this new field to a migration for version 1.3.0 so users don't have any trouble updating :)

saminakh:
@alexandreaco I updated the latest address logic to track on a per-network basis, and added some defaults for those values in a migration for v1.3.0. I'm not sure if the migration changes are what you were looking for though, let me know if I missed the mark on that and need to change it!

src/actions/electron/data-store.ts
```diff
@@ -16,6 +16,15 @@ export const saveAccountName = (event: IpcMainInvokeEvent, data: string) => {
   return store.set(`account.${accountAddress}`, prettyName)
 }
 
+export const getLatestAccountAddress = (event: IpcMainInvokeEvent, network: string): string => {
+  return store.get(`latestAddress.${network}`, '') as string
```
alexandreaco, line 20: Sorry for the confusion. We actually want to nest these under the existing STOKENET and MAINNET sections in wallet.json
Screen Shot 2021-12-01 at 9 32 02 AM

saminakh: Oh i see! I can make that change quick :) Thank you for clarifying alex!

saminakh: Should be fixed now!

src/electron-store/migrations.ts
```diff
@@ -29,6 +29,12 @@ const migrations = {
     // Remove old configuration
     store.delete('derivedAccountsIndex')
     store.delete('hardwareAddress')
+  },
+  '1.3.0': (store: Record<string, any>) => {
+    if (!store.get('wallets.MAINNET.latestAddress'))
```
alexandreaco, line 34: Last thing, you just need to update the migration to reference `wallets.${network}.latestAddress`

saminakh: I'm confused, is what is here not enough? There shouldn't be a value already for `wallets.MAINNET.latestAddress` or `wallets.STOKENET.latestAddress`, so I gave them a default of an empty string. The app won't attempt to change to an account unless it's _not_ an empty string, and this value will get set the first time the user switches to a different account.

alexandreaco: So strange! I must have had a cached old version of the PR open in github when I left that comment. You're right. When I refresh I see the correct thing. My apologies!

alexandreaco (approved):
When you rebase and merge, please include both your latestAccount and the hiddenTokens logic within the 1.3.0 migration :)

src/actions/electron/data-store.ts
```diff
@@ -16,6 +16,15 @@ export const saveAccountName = (event: IpcMainInvokeEvent, data: string) => {
...
 
+export const getLatestAccountAddress = (event: IpcMainInvokeEvent, network: string): string => {
+  return store.get(`wallets.${network}.latestAddress`, '') as string
+}
+
+export const saveLatestAccountAddress = (event: IpcMainInvokeEvent, data: string) => {
+  const { address, network } = JSON.parse(data)
+  return store.set(`wallets.${network}.latestAddress`, address)
+}
+
```
SK-Sam, line 27: oh this is great, seems like Latest Account is requiring the network as well 👐

SK-Sam (approved):
This is great! Much nicer than i imagined the solution would look like 😅 Thanks for making your work easy to follow, I'm not exposed to many of these concepts BUT have learned so much just from reading your code and feel I learned a lot about Vue/Electron 🙇

PR #397, opened 2021-12-01
chore: 4.0 SDK update

alexandreaco:
This is currently a first pass at updating the codebase to work with the latest 4.0 SDK release. 

- [x] Updates the shape of responses to reflect new types
- [x] Fixes some errors where DecodedResponse types were used instead of previous types. **note: We should double check that these responses still work when we connect to the active network. Some of these types may need to be exported from the SDK to work properly**
- [x] Fix the more logic heavy compiler errors that remain 
- [x] Connect to network and confirm the behavior works as expected

Update:

Here is a list of known logic updates that we need to make to correctly integrate with the new API
- [x] The shape of stake and unstake transactions are now different.
- [x] tokenBalancesForAddress now returns liquid accounts, stakes, and unstakes. We need to update how we display current tokens on the overview screen
- [x] tokenBalancesForAddress no longer returns full Token Info. We need to query and cache token info for each of the user's token types to compute granularity when verifying amounts
- [x] Update staking to include new pending stakes info from API
- [x] Include new robust error handling


src/components/StakeListItem.vue
```diff
@@ -141,7 +141,7 @@ const StakeListItem = defineComponent({
     const preloadedValidator = toRef(props, 'preloadedValidator')
 
     // If validator was included in top 100, use the preloaded value instead of re-fetching
-    if (!preloadedValidator.value) firstValueFrom(radix.ledger.lookupValidator(position.value.validator)).then((validatorRes: Validator) => { validator.value = validatorRes })
+    if (!preloadedValidator.value) firstValueFrom(radix.ledger.lookupValidator(position.value.validator)).then((validatorRes: any) => { validator.value = validatorRes })
```
alexandreaco, line 144: @AbstractFruitFactory I couldn't find `ValidatorEndpoint.DecodedResponse` exported from the @radix/application

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -198,7 +198,7 @@ const WalletConfirmTransactionModal = defineComponent({
       mergeMap((account) =>
         radix.ledger.tokenBalancesForAddress(account.address)
       )
-    ).subscribe((balances) => {
+    ).subscribe((balances: any) => {
```
alexandreaco, line 201: This function isn't typed in radixCoreAPI.ts @AbstractFruitFactory what should the return type be here?

src/views/Wallet/WalletOverview.vue
```diff
@@ -119,7 +119,7 @@ const WalletOverview = defineComponent({
         radix.ledger.stakesForAddress(account.address),
         radix.ledger.unstakesForAddress(account.address)
       ]))
-    ).subscribe(([balances, stakes, unstakes]) => {
+    ).subscribe(([balances, stakes, unstakes]: [any, any, any]) => {
```
alexandreaco, line 122: @AbstractFruitFactory I _think_ we need to type these with the new api types like `UnstakePositionsEndpoint.DecodedResponse` but when I do that, I get a bunch of cascading compiler errors where we reference those types. How do you imagine the client should interact with these new DecodedResponse api types?

src/components/StakeListItem.vue
AbstractFruitFactory: you probably don't need to be explicit with the type here? Would it be enough to simply let typescript infer the type?

src/composables/useTokenBalances.ts
```diff
@@ -1,27 +1,54 @@
...
-export default function useTokenBalances (radix: RadixT): useTokenBalancesInterface {
-  const tokenBalances: Ref<TokenBalances | null> = ref(null)
-  const tokenBalancesSub = radix.tokenBalances.subscribe((tokenBalancesRes: TokenBalances) => {
-    tokenBalances.value = tokenBalancesRes
-  })
+      // Get token info for tokens related to this account and save in memory
+      // Don't save duplicates
+      tokenBalancesRes.account_balances.liquid_balances.map((balance) => {
+        if (!relatedTokens.value.find((t) => t.rri.equals(balance.token_identifier.rri))) {
+          console.log('getting info for', balance)
```
saminakh, line 21: I know this is a draft, but just wanted to note this `console.log` so it doesn't get left in once it's converted to RFR! I just noticed it while doing some testing.

staylorwr (approved):
🚀

PR #400, opened 2021-12-03
fix: add missing password field to pin reset page

saminakh:
This PR updates a password field in the pin reset page.

Screen Shot 2021-12-03 at 12 43 59 PM


alexandreaco (approved):
💯

PR #405, opened 2021-12-07
Samuel/rdx 221 add additional information to the wallet

SK-Sam:
DRAFT PR(styling needs to be re-done. Began w/ `AppModal`  but will convert away from using `AppModal` due to how it looks.

[Demo](https://www.loom.com/share/fe3227396ec94ffabd64e654a8734bfd)

This PR adds an extra step between Pin confirmation and Wallet creation in the Restore Wallet flow. I've split up `handleCreatePin` logic as it seemed like it was constraining to use this function as the ending to the process. `handleCreatePin` now stores PIN upon confirmation. The actual wallet creation is being done in the interstitial screen where the user hits `I Understand`.

In this demo video, the first pass runs through the scenario where a perfect User inputs everything as planned. The next few passes run through a User who exits the app after confirming their PIN, just to show that wallet creation is dependent on the user hitting `I Understand`. If you see nothing happening, it's me inputting my seed phrase.

TODO:
- [ ] move all text to `text/index.ts` to be i18n-able
- [ ] change style so interstitial screen doesn't look like popup modal
- [ ] confirm that logic is correct

### DIFF
--- a/src/views/RestoreWallet/index.vue
+++ b/src/views/RestoreWallet/index.vue
@@ -71,6 +71,7 @@
         @enteredPin="handleEnterPin"
       >
       </create-wallet-create-pin>
+      <multiple-accounts-disclaimer-modal/>
     </div>
   </div>
 </template>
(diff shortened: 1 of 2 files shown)


src/views/RestoreWallet/index.vue
```diff
@@ -136,29 +141,37 @@ const RestoreWallet = defineComponent({
...
-        setNetwork(network)
-        walletLoaded()
-        return waitUntilAllLoaded()
-      }).then(() => {
-        setState(true)
-        router.push('/wallet/account-edit-name')
-      })
+    }
+
+    const completeWalletRestore = (isUnderstood: boolean) => {
```
saminakh, line 150: Pulling this logic out of the `handleCreatePin` function was a good choice imo!

src/views/RestoreWallet/index.vue
```diff
@@ -66,12 +66,16 @@
       </create-wallet-create-passcode>
 
       <create-wallet-create-pin
-        v-if="step == 2 || step == 3"
+        v-if="step == 2 || step == 3 && !pinIsSet"
```
saminakh, line 69: With the move away from a modal, would it make sense to treat the "i understand" screen as a step 4 and make the `v-if`s here a little more uniform?

saminakh (commented):
I think this is looking good so far! I'll take another look when it's no longer a draft 👍

PR #407, opened 2021-12-08
chore: add legal terms to home view

saminakh:
This PR adds the TOS to the home screen, displaying it if it has not been accepted yet. Once it's been accepted, the value is added to the store; if the terms ever need to change, we should be able to add a migration to reset that field to 'false' for all users again, forcing them to accept the new terms before use.

Below are a few screenshots of the implementation:
Screen Shot 2021-12-07 at 8 08 19 PM
Screen Shot 2021-12-07 at 8 08 31 PM
Screen Shot 2021-12-07 at 8 08 38 PM


caleboller (approved):
@saminakh! I was wondering how we'd go about rendering the ToS. I would not have guessed we'd just hand write 'em, but good on you! 

This looks beautiful and works 💯

PR #410, opened 2021-12-21
chore: api error handling

saminakh:
This PR adds an additional error modal specifically for handling API errors. For known errors, a friendly message is displayed; otherwise, a detailed modal with the error message, trace id, and contact email is displayed.

Examples:
Screen Shot 2022-01-07 at 3 17 22 PM
Screen Shot 2022-01-07 at 3 18 33 PM
Screen Shot 2022-01-06 at 11 40 04 PM

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -57,7 +57,7 @@ const App = defineComponent({
     const isApiError: ComputedRef<boolean> = computed(() => {
       const latest: any = appErrors.value[appErrors.value.length - 1]
       if (latest === null) return false
-      return latest.error?.details?.type
+      return latest.type === 'api'
     })
 
     const errorsCount: ComputedRef<number> = computed(() => appErrors.value.length)
--- a/src/App.vue
+++ b/src/App.vue
@@ -31,7 +31,6 @@ import { useWallet, useErrors } from './composables'
 import ErrorModalGeneric from '@/components/ErrorModalGeneric.vue'
 import ErrorModalTransactionBuilding from '@/components/ErrorModalTransactionBuilding.vue'
 import ErrorModalTransactionConfirming from '@/components/ErrorModalTransactionConfirming.vue'
-import { ClientAppErrorT } from './composables/useErrors'
 
 const App = defineComponent({
   components: {
--- a/src/App.vue
+++ b/src/App.vue
@@ -3,8 +3,13 @@
     <div class="bg-rBlue w-1/2 h-tallest origin-center transform rotate-45 absolute z-0 top-0 left-0 -translate-x-1/2 -translate-y-1/2"></div>
     <div class="relative z-10 min-h-screen">
       <template v-if="latestError">
+        <ErrorModalApi
+          v-if="isApiError"
+          :error="latestError.error"
+          :errorsCount="errorsCount"
+        />
         <ErrorModalTransactionBuilding
-          v-if="latestError.type === 'TRANSACTION-BUILDING'"
+          v-else-if="latestError.type === 'TRANSACTION-BUILDING'"
           :error="latestError.error"
           :errorsCount="errorsCount"
         />
@@ -29,6 +34,7 @@ import { computed, ComputedRef, defineComponent } from 'vue'
 import { useRouter } from 'vue-router'
 import { useWallet, useErrors } from './composables'
 import ErrorModalGeneric from '@/components/ErrorModalGeneric.vue'
+import ErrorModalApi from '@/components/ErrorModalApi.vue'
 import ErrorModalTransactionBuilding from '@/components/ErrorModalTransactionBuilding.vue'
 import ErrorModalTransactionConfirming from '@/components/ErrorModalTransactionConfirming.vue'
 
@@ -36,7 +42,8 @@ const App = defineComponent({
   components: {
     ErrorModalGeneric,
     ErrorModalTransactionBuilding,
-    ErrorModalTransactionConfirming
+    ErrorModalTransactionConfirming,
+    ErrorModalApi
   },
 
   setup () {
@@ -47,12 +54,19 @@ const App = defineComponent({
     const latestError: ComputedRef<Error | null> = computed(() => {
       return appErrors.value[appErrors.value.length - 1]
     })
+    const isApiError: ComputedRef<boolean> = computed(() => {
+      const latest: any = appErrors.value[appErrors.value.length - 1]
+      if (latest === null) return false
+      return latest.error?.details?.type
+    })
+
     const errorsCount: ComputedRef<number> = computed(() => appErrors.value.length)
 
     return {
       appErrors,
       errorsCount,
-      latestError
+      latestError,
+      isApiError
     }
   }
 })
--- a/src/components/ErrorModalApi.vue
+++ b/src/components/ErrorModalApi.vue
@@ -75,7 +75,7 @@ export default defineComponent({
     const error = toRef(props, 'error')
     const router = useRouter()
     const { radix } = useWallet(router)
-    const { clearLatestError, isApiError } = useErrors(radix)
+    const { clearLatestError, isKnownApiError } = useErrors(radix)
     const { t } = useI18n()
 
     watch((error), (newErrorVal) => {
@@ -89,7 +89,7 @@ export default defineComponent({
       if (props.errorsCount <= 1) isVisible.value = false
     }
     const type: string = error.value?.details?.type
-    const showDetails = !isApiError(type)
+    const showDetails = !isKnownApiError(type)
     let errorMsg: string = t('apiErrors.unknown')
     let errorTitle: string = t('apiErrors.unknown')
     if (!showDetails) {
--- a/src/components/ErrorModalApi.vue
+++ b/src/components/ErrorModalApi.vue
@@ -15,9 +15,17 @@
         <div class="bg-rRed rounded-full inline-block px-2 py-0.5 m-2 text-white text-sm">{{ errorsCount }}</div>
       </div>
       <div v-if="showDetails">
-        <p class="mb-2">Error Code: {{ error.code }}</p>
-        <p class="mb-2">Error Type: {{ error.details.type }}</p>
-        <p class="mb-5">{{ error.message }}</p>
+        <p class="mb-2">Please contact support at <a class="text-rBlue"
+        href="mailto:hello@radixdlt.com">hello@radixdlt.com</a> and provide the following.
+        </p>
+        <p class="mb-2 flex justify-between">
+          <span class="flex-1">Error: {{ error.details.type }}</span>
+          <click-to-copy :address="error.details.type" />
+        </p>
+        <p class="mb-5 flex justify-between">
+          <span class="flex-1">Trace ID: {{ error.trace_id }}</span>
+          <click-to-copy :address="error.trace_id" />
+        </p>
       </div>
       <div class="flex flex-row space-x-5 justify-center">
         <AppButtonCancel @click="handleClose" class="w-44">{{ $t('errors.closeModal') }}</AppButtonCancel>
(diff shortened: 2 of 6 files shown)


alexandreaco:
I can't leave a review because it's still a WIP branch, but here's some feedback :) @saminakh 

1. https://github.com/radixdlt/olympia-wallet/pull/410/files#diff-7a7a37b12ee1265d7e27577ec286120d2cbc0940908635e264a2be44ccb9a8a0R6-R13 I don't think we'll ever show the transaction building error because isApiError will return true first
2. https://github.com/radixdlt/olympia-wallet/pull/410/files#diff-7deb4a2ae75ee9c0d8e9807092fd413c257fae59ca15a80bb16276daf099577eL244-R244 Typo, we lost a space between words
3. I wonder if we want the copy to say "Unknown Api Error" or something more confident like "Received Api Error" when we don't have friendly text to show. Screen Shot 2022-01-05 at 9 13 00 AM 
4. I'm having trouble creating a "not enough XRD to cover fee" error. This may be an issue with the Api Screen Shot 2022-01-05 at 9 14 53 AM 
5. For `NotEnoughTokensForUnstakeError`, we should check with Radix to see if we need to present both of those messages to the user

saminakh:
Thank you for the feedback @alexandreaco !
1. Todo
2. Fixed!
3. I think that's a good idea! I changed it!
4. I did manage to reproduce this, it appears that the API only sends this error if you're sending NON-native tokens and you don't have enough native tokens to cover the transaction fee. There doesn't appear to be a corresponding error for if you're sending native tokens though and don't have enough to cover the transaction fee. Should we tell radix and recommend they add an api error for that?
5. Todo

src/components/ErrorModalApi.vue
```diff
@@ -0,0 +1,120 @@
...
+      clearLatestError()
+      // Clear latest error but leave modal open if there are more errors in the list
+      // Close modal if this is the last error
+      if (props.errorsCount <= 1) isVisible.value = false
+    }
+    const type: string = error.value?.details?.type
+    const showDetails = !isApiError(type)
+    let errorMsg: string
+    if (showDetails) {
+      errorMsg = 'Unexpected Error'
```
alexandreaco, line 92: Let's make sure we're using the i18n text file for these bits of text :)

src/components/ErrorModalApi.vue
```diff
@@ -0,0 +1,120 @@
...
+    } else if (type === 'BelowMinimumStakeError') {
+      const minimum = error.value.details?.minimum_amount?.value
+      const displayMinimum = minimum ? asBigNumber(minimum) : '10'
+      errorMsg = t(`apiErrors.${type}`, { minimum: displayMinimum })
```
alexandreaco, line 96: nit: I don't think you need to template the string `apiErrors.${type}` because we always know the type is `BelowMinimumStakeError`. Nice use of custom values in the i18n call!

src/components/ErrorModalApi.vue
```diff
@@ -0,0 +1,120 @@
...
+      const requested = Amount.fromUnsafe(error.value.details.requested_amount.value)._unsafeUnwrap()
+      const staked = Amount.fromUnsafe(error.value.details.stake.delegated_stake.value)._unsafeUnwrap()
+      const pending = Amount.fromUnsafe(error.value.details.pending_stake.delegated_stake.value)._unsafeUnwrap()
+      if (requested > add(staked, pending)) {
+        errorMsg = t(`apiErrors.${type}.default`)
+      } else {
+        errorMsg = t(`apiErrors.${type}.pending`, { pending: asBigNumber(pending) })
+      }
+    } else {
+      errorMsg = t(`apiErrors.${type}`)
```
alexandreaco, line 107: Would we really ever get to this clause? Maybe you can reverse your conditionals and just leave a single "else"

if (requested <= add(staked, pending))
  // show pending error
else 
  // show default error

src/composables/useTransactions.ts
```diff
@@ -269,8 +273,12 @@ export default function useTransactions (radix: ReturnType<typeof Radix.create>,
     userDidConfirm.next(false)
     shouldShowConfirmation.value = true
 
-    transactionSubs.add(completion.subscribe(handleTransactionCompleted))
-    transactionSubs.add(events.subscribe(handleTransactionLifecycleEvent))
+    transactionSubs.add(
+      completion.subscribe(handleTransactionCompleted, errorHandler)
```
alexandreaco, line 277: neat! i didn't know you could pass multiple functions into subscribe

src/composables/useWallet.ts
```diff
@@ -85,7 +85,7 @@ const createWallet = (mnemonic: MnemomicT, pass: string, network: Network) => {
     hasWallet.value = true
     setWallet(newWallet)
     saveDerivedAccountsIndex(0, network)
-    persistNodeUrl('https://sandpitnet-gateway.radixdlt.com')
+    persistNodeUrl('https://stokenet-gateway.radixdlt.com')
```
alexandreaco, line 88: I think you'll have rebase conflicts because I made these strings a const. sorry!

src/text/index.ts
```diff
@@ -242,6 +242,22 @@ const messages = {
       unableToPrepareUnstakingTransactionPointOne: 'Attempting to request an unstake in the same “epoch” (~30 min period) that you staked',
       unableToPrepareUnstakingTransactionPointTwo: 'Attempting to unstake more than you currently have staked to the validator',
       unableToPrepareUnstakingTransactionPointThree: 'Insufficient XRD to pay the required transaction fee'
+    },
+    apiErrors: {
+      NetworkNotSupportedError: 'The network selected is not valid.',
```
alexandreaco, line 247: We can remove some of these errors. We should only show custom errors for the following codes

InvalidAccountAddressError
InvalidValidatorAddressError
NotEnoughNativeTokensForFeesError
NotEnoughTokensForTransferError
NotEnoughTokensForStakeError
NotEnoughTokensForUnstakeError
BelowMinimumStakeError
CannotStakeError
MessageTooLongError

src/composables/useErrors.ts
```diff
@@ -1,20 +1,27 @@
...
+export const apiErrors = [
+  'NetworkNotSupportedError',
+  'InvalidTokenRRIError',
+  'InvalidAccountAddressError',
+  'InvalidValidatorAddressError',
+  'NotEnoughNativeTokensForFeesError',
+  'NotEnoughTokensForTransferError',
+  'NotEnoughTokensForStakeError',
+  'NotEnoughTokensForUnstakeError',
+  'BelowMinimumStakeError',
```
alexandreaco, line 15: We also need to remove the inline check for 90xrd

src/composables/useErrors.ts
```diff
@@ -1,20 +1,27 @@
...
+  'CannotStakeError',
+  'MessageTooLongError'
```
alexandreaco, line 17: We also need to remove the inline check for when messages are longer than 160 characters

alexandreaco (changes requested):
Thanks for all of this work! 🙇‍♀️  I left a few notes about small nits and minor updates

src/composables/useWallet.ts
saminakh: No worries! Rebased 👍

src/components/ErrorModalApi.vue
saminakh: Oh! I think i named the error messages confusingly. I'll change this to make it clearer!

src/components/ErrorModalApi.vue
saminakh: 👍

src/components/ErrorModalApi.vue
saminakh: Yeah good point! I did this in another spot too so I'll change 'em both!

src/composables/useTransactions.ts
saminakh: I didn't either! someone from radix had a code example in the channel 😄

src/text/index.ts
saminakh: 👍 totally forgot to remove these cases, thanks for the reminder!

src/composables/useErrors.ts
saminakh: I should have asked about the inline validations, that was silly of me not to. I thought maybe these modals would just be backups if the inline checks failed? But that was a silly assumption in retrospect 😬

src/composables/useErrors.ts
saminakh: 👍

saminakh:
@alexandreaco for title/copy change for the error modals, I went with:
- Cannot Create Transaction
- Cannot Complete Stake
- Cannot Complete Unstake
Do those sound reasonable?

src/App.vue
```diff
@@ -3,8 +3,13 @@
     <div class="bg-rBlue w-1/2 h-tallest origin-center transform rotate-45 absolute z-0 top-0 left-0 -translate-x-1/2 -translate-y-1/2"></div>
     <div class="relative z-10 min-h-screen">
       <template v-if="latestError">
+        <ErrorModalApi
+          v-if="isApiError"
```
alexandreaco, line 7: `isApiError` returns true if `latest.error?.details?.type` exists. This means that we'll never get to the subsequent else-if statements that are checking for that type to be a specific string. Maybe you intended to import `isApiError` from `useErrors`?

src/components/ErrorModalApi.vue
```diff
@@ -0,0 +1,135 @@
...
+      <path d="M25 14V36" stroke="#052CC0" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+      <path d="M14 25H36" stroke="#052CC0" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
+      </svg>
+    </template>
+    <template v-slot:content>
+      <div v-if="errorsCount > 1" class="absolute top-0 right-0">
+        <div class="bg-rRed rounded-full inline-block px-2 py-0.5 m-2 text-white text-sm">{{ errorsCount }}</div>
+      </div>
+      <div v-if="showDetails">
+        <p class="mb-2">Please contact support at <a class="text-rBlue"
```
alexandreaco, line 18: Can we add these phrases to the text.ts file?

src/text/index.ts
```diff
@@ -242,6 +242,51 @@ const messages = {
       unableToPrepareUnstakingTransactionPointOne: 'Attempting to request an unstake in the same “epoch” (~30 min period) that you staked',
       unableToPrepareUnstakingTransactionPointTwo: 'Attempting to unstake more than you currently have staked to the validator',
       unableToPrepareUnstakingTransactionPointThree: 'Insufficient XRD to pay the required transaction fee'
+    },
+    apiErrors: {
+      unknown: 'Unexpected Error',
+      InvalidAccountAddressError: {
+        message: 'The recipient Radix address specified is not valid.',
```
alexandreaco, line 249: 👍 This nesting is really easy to follow and documents the flow well. Great idea

src/views/Wallet/WalletTransaction.vue
```diff
@@ -96,9 +96,6 @@
                       class="w-full"
                       label="Message"
                       :placeholder="$t('transaction.messagePlaceholder')"
-                      :rules="{
```
alexandreaco, line 99: I'm seeing a runtime error now that rules is undefined. Can you set rules to an empty object instead?

alexandreaco (changes requested):
The new error modals look great. Thanks for adding in the titles :) 

I left a few notes in the PR that we should cleanup before merging. 

Lastly, can you add some horizontal padding to this sentence so we don't have an orphaned word on the second line? In the screenshot I tried px-12 and it looks nicer.
Screen Shot 2022-01-10 at 6 29 58 AM

src/App.vue
saminakh: Ahh I see what's confusing here. The flow should be: check that `latest.error.details.type` exists. If yes: show api error modal, where we look at the type (and if the type is found, show that simplified error modal for that type, otherwise, show the detailed error modal w/ the trace id). If `latest.error.details.type` doesn't exist, _then_ we think it's not an API error and show one of the other modals instead.

So I think the problem here is, is there a better way to check if it's an API error or not? I can look a little more closely at the error and see if there's a nicer way to do this.

src/components/ErrorModalApi.vue
saminakh: Yeah! Should i just break the string into two chunks surrounding the link?

src/text/index.ts
saminakh: 😄

src/views/Wallet/WalletTransaction.vue
saminakh: Good catch! Sorry about that, I'll go and fix it!

src/components/ErrorModalApi.vue
saminakh: (I ended up just doing that!)

src/App.vue
saminakh: Ok! Here's what I did. I:
- updated where we handle the api errors to include a `type` flag of 'api' before adding it to the list of errors for the app. This 1) is probably a better way of checking if the error is an api type or a general app error, and 2) makes the logic a little more clear
- renamed the `isApiError` --> `isKnownApiError`, to make the purpose of that function more clear

alexandreaco (approved):
💯

src/App.vue
alexandreaco: Thanks, this makes much more sense now!

PR #412, opened 2022-01-06
chore: Upgrade to 4.0.9 for SDK fixes

alexandreaco:
Includes... 
- fixes decrypting old messages in the transaction history
- adds @radixdlt/crypto to correctly parse unencrypted messages in history


SK-Sam (approved):
Hey! I see my older messages 😄 

Screen Shot 2022-01-06 at 10 17 05 AM

saminakh:
> Hey! I see my older messages 😄
> 
> Screen Shot 2022-01-06 at 10 17 05 AM

One of those transactions looks familiar 👀

src/composables/useTokenBalances.ts
```diff
@@ -18,7 +18,6 @@ export default function useTokenBalances (radix: ReturnType<typeof Radix.create>
       // Don't save duplicates
       tokenBalancesRes.account_balances.liquid_balances.map((balance) => {
         if (!relatedTokens.value.find((t) => t.rri.equals(balance.token_identifier.rri))) {
-          console.log('getting info for', balance)
```
saminakh, line 21: 🧹

src/composables/useWallet.ts
```diff
@@ -85,7 +85,7 @@ const createWallet = (mnemonic: MnemomicT, pass: string, network: Network) => {
     hasWallet.value = true
     setWallet(newWallet)
     saveDerivedAccountsIndex(0, network)
-    persistNodeUrl('https://sandpitnet-gateway.radixdlt.com')
+    persistNodeUrl('https://stokenet-gateway.radixdlt.com')
```
saminakh, line 88: Would it be easier to define a default node somewhere once and then retrieve it in all these places, so you don't have to keep writing the same url multiple times?

saminakh (approved):
Code LGTM! And appears to be having as expected locally for me!

src/composables/useWallet.ts
alexandreaco: That's a great idea!

PR #413, opened 2022-01-10
Improve display of complex actions in history

alexandreaco:
This PR makes some small UI updates to how we display edge cases actions in the history view.

1. For transactions with an empty actions array, we display "Unknown transaction." This most often occurs when we return legacy self-transfers
Screen Shot 2022-01-10 at 6 15 22 AM

2. For transactions with multiple actions, i.e. complex transactions, we display "Complex transaction" and tell the user if some of the actions were filtered out because they don't relate to the current active address
Screen Shot 2022-01-10 at 6 15 13 AM

See notes in linear for reference https://linear.app/township/issue/RDX-314/sdk-upgrade-history


staylorwr (approved):
Lovely, thank you Alex.

PR #415, opened 2022-01-13
fix: Correctly label validators that are not in top 100

alexandreaco:


SK-Sam (changes requested):
👋 I've stubbed some false values to force every Validator to not be in the top 100 and the `span`s are showing up just fine! 🥳 

One nit I found while growing my Olympia screen to max size, some `span`s are showing up on the same line as the previous `span`, IE: Unregistered and Not in top 100 sometimes share the same line. Here's a screenshot
Screen Shot 2022-01-13 at 9 50 31 AM

If I were to change L16 by turning `inline-flex` to `flex`, it seems to do the trick.
`<div v-if="!inTopOneHundred" class="flex items-center">`
Screen Shot 2022-01-13 at 9 54 15 AM

alexandreaco:
Thanks for the catch @SK-Sam, I updated it

staylorwr (approved):
Hard to validate the validator logic but code looks good 👍🏼

SK-Sam (approved):
LGTM 😄 Thanks for the changes!

PR #418, opened 2022-01-14
chore: Surface simple error when user disconnects hardware device mid-transaction

alexandreaco:
https://www.loom.com/share/a698910ebb7841c5ab7be1c8abaa0bfd

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -8,6 +8,11 @@
           :error="latestError.error"
           :errorsCount="errorsCount"
         />
+        <ErrorModalHardware
+          v-if="latestError.type === 'HARDWARE'"
+          :error="latestError.error"
+          :errorsCount="errorsCount"
+        />
         <ErrorModalTransactionBuilding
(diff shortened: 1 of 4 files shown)


SK-Sam:
I tried the fixes and it works for the flow of Disconnecting and THEN starting a transaction

Do we need to take into account when a User unplugs WHILE the transaction is waiting for Ledger to review/sign? When I am reviewing details on Ledger, I unplug and am stuck at a `Submitting Transaction...Confirmed` state.
Screen Shot 2022-01-14 at 12 13 19 PM

SK-Sam:
Here are the errors that surface when I unplug Ledger while reviewing tx/signing tx...
Screen Shot 2022-01-14 at 12 17 18 PM

alexandreaco:
Thanks @SK-Sam! I took a look and added another case to the logic. Can you test this now? It should catch both cases

staylorwr (approved):
Works well!

PR #419, opened 2022-01-14
fix: Tell user to confirm on ledger device at all relevant steps

alexandreaco:
This PR fixes a bug where not all relevant transaction states were showing the "confirm and sign" modal when sending an encrypted message. it also fixes a bug where you could end up with multiple modals because we were missing if-else statements.

### DIFF
--- a/src/views/Wallet/WalletConfirmTransactionModal.vue
+++ b/src/views/Wallet/WalletConfirmTransactionModal.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="fixed w-screen h-screen z-20 flex items-center justify-center bg-translucent-black">
     <div class="h-modalMedium bg-white rounded-md py-7 px-7 w-full max-w-3xl absolute top-1/2 left-1/2 transform -translate-x-1/3 -translate-y-1/2">
-      <div v-if="transactionState === 'INITIATED' || transactionState === 'BUILT_FROM_INTENT'">
+      <div v-if="shouldShowBuildingModal">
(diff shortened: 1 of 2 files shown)


src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -33,7 +33,7 @@
         </div>
       </div>
       <form
-        v-if="transactionState === 'ASKED_FOR_CONFIRMATION'"
+        v-else-if="transactionState === 'ASKED_FOR_CONFIRMATION'"
```
SK-Sam, line 36: gotcha!

SK-Sam (approved):
This works well 😄 Tested locally for both Staking & Sending.

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -20,7 +20,7 @@
           </button>
         </div>
       </div>
-      <div v-if="transactionState === 'hw-signing'">
+      <div v-else-if="transactionState === 'hw-signing' || transactionState === 'confirm' || transactionState === 'CONFIRMED'">
```
staylorwr, line 23: could we turn this into a computed property?

staylorwr (approved):
Small nit about multi-case if statements but everything else looks 💯

src/views/Wallet/WalletConfirmTransactionModal.vue
alexandreaco: Great catch, thanks!

PR #422, opened 2022-01-19
chore: Upgrade SDK fix to include error when switching to invalid gateway

alexandreaco:
Screen Shot 2022-01-19 at 9 43 57 AM


staylorwr (approved):
Looks great!

PR #423, opened 2022-01-20
fix: Default to Mainnet for new wallets

staylorwr:
New Wallets or wallets without a custom node url should default to mainnet,  we had changed this value while the SDK / API were being updated.  This can be restored to it's original value now that the mainnet gateway is live.


SK-Sam (approved):
Can confirm it is indeed defaulting to Mainnet!

PR #424, opened 2022-01-21
fix: network versions

staylorwr:
This PR makes two small changes:
* The network helpers used for the default list of available gateways was incorrect
* The case of the string representing the activeNetwork has changed in the latest SDK so the test network warning was showing even when connected to mainnet.

### DIFF
--- a/src/views/Wallet/index.vue
+++ b/src/views/Wallet/index.vue
@@ -5,7 +5,7 @@
     <div class="flex flex-col w-full">
       <router-view v-if="walletLoaded"/>
       <wallet-loading v-else />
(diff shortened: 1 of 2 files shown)


SK-Sam (approved):
Screen Shot 2022-01-21 at 1 50 51 PM

👍

src/views/Wallet/index.vue
```diff
@@ -82,13 +82,18 @@ const WalletIndex = defineComponent({
       router.push({ name: 'Home', query: { modal: '' } })
     })
 
+    const isTestNet = computed(() => {
+      return activeNetwork.value && activeNetwork.value !== 'mainnet'
```
saminakh, line 86: I like this!

saminakh (approved):

PR #426, opened 2022-01-24
fix: Correctly show hardware error modal

alexandreaco:
Screen Shot 2022-01-24 at 9 16 36 AM

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -9,7 +9,7 @@
           :errorsCount="errorsCount"
         />
         <ErrorModalHardware
-          v-if="latestError.type === 'HARDWARE'"
+          v-else-if="latestError.type === 'HARDWARE'"
           :error="latestError.error"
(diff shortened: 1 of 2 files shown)


staylorwr (approved):
Functions as expected :100:

SK-Sam:
👋 Stress testing other facets of message length and inputting past acceptable char limits
- When I input a message in transaction normally for BOTH non-hardware & hardware tx(no encryption, > char limit) this works well ✅ 
- When encrypting a msg that is too long, I run into a continuous message `"BUILT_FROM_INTENT"` and receive this error
Screen Shot 2022-01-24 at 9 58 55 AM

alexandreaco:
Thanks for finding this @SK-Sam! Another error wasn't being caught by the SDK. I'm massaging it into the API error format as a temporary patch

PR #427, opened 2022-01-25
Fix: clear caps warning on pw field on focus lost

saminakh:
This PR updates the password field to clear the caps lock warning when the field loses focus. This helps with screens where there are multiple password fields (ie password update, where you have current, new, and confirm new password fields); previously each individual field would track the caps lock status independently, resulting in some odd looking states. Now, the caps lock warning clears when you change fields and shouldnt display in multiple places.

### DIFF
--- a/src/components/PasswordField.vue
+++ b/src/components/PasswordField.vue
@@ -10,6 +10,7 @@
       :label="label"
       :validateOnInput="validateOnInput"
       :validateOnBlur="validateOnBlur"
+      @blur="clearCapsLockStatus"
       @keyup="checkCapsLock"
       @keydown.caps-lock="checkCapsLock"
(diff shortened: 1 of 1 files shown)


src/components/PasswordField.vue
```diff
@@ -10,6 +10,7 @@
       :label="label"
       :validateOnInput="validateOnInput"
       :validateOnBlur="validateOnBlur"
+      @blur="clearCapsLockStatus"
```
SK-Sam, line 13: 👍

SK-Sam (approved):
I tried to break this 😄 foolproof!! LGTM.

alexandreaco (approved):
💯

PR #428, opened 2022-01-26
chore: create parsing helper fn to parse transaction states

SK-Sam:
[Demo](https://www.loom.com/share/20d6ece30c384ac0aae9bbdb897f77a7)

This PR creates a helper function `parseUnderscoresToSpaces` and parses every `transactionState`. Previously the state `UPDATE_OF_STATUS_PENDING_TX` was displaying as is. It seems we are actually just displaying whatever comes from the API, and coincidentally this transaction state is the only multi-word state we're surfacing to the User.

Current list of transaction related messages we're receiving from API:
image


caleboller:
@SK-Sam This is a really clever solution! 👏 👏

PR #433, opened 2022-01-26
chore: Display friendly messages for transaction states

alexandreaco:


SK-Sam (approved):
Just tried it & worked perfectly, LGTM 😄 

One thing I was thinking of earlier...what if we never noticed this `UPDATE_PENDING_TX` multi-word message because it was newly added? Maybe we should at some point
1: Create a function that displays a generic message in case of a Key the API emits that we don't account for(a new transaction state key the SDK team creates)
2: Just surface with the Radix team that we're translating messages we expect from them, in order to make it easier on the User's eyes. Wondering if they are using those transaction state msgs on their end, or if they coincidentally expect we like the format with underscores 🤔

PR #434, opened 2022-01-27
fix: send user to settings page if unable to login

saminakh:
This PR fixes an issue where we weren't handling an error on login resulting from being unable to connect to custom nodes, ie https://coinwrap.co, and the error was instead being treated as an invalid password error. The user should now get sent to the settings page with an error modal if they provide a valid password but are unable to connect to their current custom node.

https://www.loom.com/share/d4b83ef722394ead95a9757a9f55fa80

### DIFF
--- a/src/components/ErrorModalGeneric.vue
+++ b/src/components/ErrorModalGeneric.vue
@@ -14,7 +14,7 @@
       <div v-if="errorsCount > 1" class="absolute top-0 right-0">
         <div class="bg-rRed rounded-full inline-block px-2 py-0.5 m-2 text-white text-sm">{{ errorsCount }}</div>
       </div>
-      <p class="mb-5">{{ error }}</p>
+      <p class="whitespace-pre-line mb-5">{{ errorMessage }}</p>
       <div class="flex flex-row space-x-5 justify-center">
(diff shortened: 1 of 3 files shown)


src/components/ErrorModalGeneric.vue
```diff
@@ -14,7 +14,7 @@
       <div v-if="errorsCount > 1" class="absolute top-0 right-0">
         <div class="bg-rRed rounded-full inline-block px-2 py-0.5 m-2 text-white text-sm">{{ errorsCount }}</div>
       </div>
-      <p class="mb-5">{{ error }}</p>
+      <p class="whitespace-pre-line mb-5">{{ errorMessage }}</p>
```
caleboller, line 17: @saminakh I just want to make sure this variable change doesn't break any other uses of this modal. hae you confirmed?

saminakh: Yeah! I tested it out with the existing error modal presented when you disconnect after logging in (rather than before). Previously, this error modal showed some ugly quotes. This cleans that up!

caleboller (approved):
Looks great! Saw a Loom video proving it works. :)

PR #437, opened 2022-01-28
Only show hardware modal when in hardware transaction flow

alexandreaco:
The old code had a few of bugs.

1. it was conflating the transactionState.value === 'CONFIRMING' and when the ledger should be signing.
    - Fixed this by tracking ledger state separately
2. it was not clearing transactionState after canceling and completing transactions.
    - Fixed this and caught a second bug where denying a transaction on the ledger device wasn't closing the modal
3. it was initializing the transactionState as 'confirm' and mistakenly using that status to mean that the user was asked to encypt on the ledger device before initializing the transaction
    - There is actually no response from the SDK that tells the client that the user has been asked to encrypt a hw transaction before initializing. In this new solution we are just showing the "pending" state in the building modal. We need more info from the SDK to round this out.

There is a `To Do` in the code where we need to update the `shouldShowLedgerModal` value to also show the ledger modal when we ask the user to encrypt a hw message early


staylorwr (approved):

staylorwr:
This is one of those things that I expect will get a major cleaning up when the new SDK updates hit in the near future 👍🏼

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -278,8 +279,10 @@ const WalletConfirmTransactionModal = defineComponent({
       transactionState.value === 'INITIATED' || transactionState.value === 'BUILT_FROM_INTENT'
     )
 
+    // Show ledger modal when hw-signing is in progress and the transaction is CONFIRMED
+    // To Do: Need a way to show modal to user who needs to encrypt a hardware message before transaction is INITIATED
     const shouldShowLedgerModal: ComputedRef<boolean> = computed(() =>
-      transactionState.value === 'hw-signing' || transactionState.value === 'confirm' || transactionState.value === 'CONFIRMED'
+      (transactionState.value === 'CONFIRMED') && ledgerState.value === 'hw-signing'
```
SK-Sam, line 285: 👍 would this be a case where we should demand for more explicit transaction states from the Radix team? (outside of the scope of this PR)

SK-Sam (approved):
Gave it a spin and works wonderfully! 😄 left a general Q for moving-forward things

PR #438, opened 2022-02-03
feat: Handle incomplete hardware wallet signature 

SK-Sam:
We were previously passing any error outside of `useTransactions.ts` pattern matching as a generic `api` type error. This includes Ledger signature rejecting & Ledger signature time out. I've added onto the logic in `useTransactions.ts` to capture these specific events along with a better way to display the error messages.
Now:
Screen Shot 2022-02-03 at 3 59 21 PM

Previously:
af73a94d-b8c3-4414-9c4f-21bd9d48fcf0


src/components/ErrorModalHardware.vue
```diff
@@ -67,8 +69,25 @@ export default defineComponent({
...
+    const errorTitle: string = t('errors.genericErrorTitle')
+    if (type) {
+      switch (type) {
+        case 'SignatureTimedOut': {
+          errorMsg = t('errors.hardwareSignatureTimedOut')
+          break
+        }
+        case 'UserRejectedSignature': {
+          errorMsg = t('errors.hardwareUserRejectedSignature')
+          break
```
saminakh, line 83: do we need a default case here? I can't recall if this code will 💩 itself if it's going through a case statement and doesn't encounter a match

saminakh: nevermind! I just googled it, it does not in fact require a default case. TIL!

src/composables/useTransactions.ts
```diff
@@ -259,6 +259,22 @@ export default function useTransactions (radix: ReturnType<typeof Radix.create>,
...
       })
+    } else if (err.toString().indexOf('Error: Failed to sign tx with Ledger') >= 0 && err.toString().includes('(0x6985')) {
+      setError({
+        ...err,
+        type: 'HARDWARE',
+        error: {
+          category: 'UserRejectedSignature'
+        }
+      })
+    } else if (err.toString().indexOf('Error: Failed to sign tx with Ledger') >= 0 && err.toString().includes('(0x530c')) {
```
saminakh, line 270: This feels like a fragile way to have to assign a type to an error. Should we make a note of this for the radix team to look at so they can give us better formed errors?

saminakh (approved):
code LGTM! Maybe worth pinging caleb or alex on the suggestion i mentioned? I don't know how feasible it is for radix to add a source or type to the errors being given here, but it would certainly be nice to have one day!

src/composables/useTransactions.ts
SK-Sam: Yeahh I was thinking about this all day yesterday. When we receive the error, it's simply a string with a unique error code. I wonder if we can beef up our own composables, or like you mentioned to request/discuss  a specific format?

src/components/ErrorModalHardware.vue
SK-Sam: I'm treating the switch logic as what transforms the msg & title 😄 FWIW I had to look up if it needs a default case as well haha

PR #439, opened 2022-02-04
fix: Isolate History Logic

staylorwr:
This PR isolates the logic for the History UI into an independent composeable that manages a single, interval driven subscription.

There have been a significant number of recent performance problems specifically with the history page.  We suspect that this may be due to rxjs subscriptions that were not completely cleaned up on route transitions or pagination logic.  

This works also replaces the `radix.transactionHistory` call with a collection of rxjs manipulations with a single api `radix.ledger.transactionHistory` call that returns a promised result on a `interval` subscription.  This should help eliminate the internal side effects of the main `radix` subscribe-ables and cleans up the responsibility of watching for changes in dependent values by isolating those responsibilities to the Vue app.

### DIFF
--- a/src/views/Wallet/WalletHistory.vue
+++ b/src/views/Wallet/WalletHistory.vue
@@ -20,9 +20,9 @@
       <div v-if="loadingHistory || !nativeToken" class="p-4 flex items-center justify-center">
         <loading-icon class="text-rGrayDark" />
       </div>
-      <template v-else-if="pendingTransactions.length > 0 || transactionsWithMessages.length > 0">
+      <template v-else-if="transactionsWithMessages.length > 0">
         <div>
-          <transaction-list-item
+          <!-- <transaction-list-item
             v-for="(txn, i) in pendingTransactions"
             :key="i"
             :transaction="txn"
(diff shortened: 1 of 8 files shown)


src/views/Wallet/WalletHistory.vue
```diff
@@ -20,9 +20,9 @@
       <div v-if="loadingHistory || !nativeToken" class="p-4 flex items-center justify-center">
         <loading-icon class="text-rGrayDark" />
       </div>
-      <template v-else-if="pendingTransactions.length > 0 || transactionsWithMessages.length > 0">
+      <template v-else-if="transactionsWithMessages.length > 0">
         <div>
-          <transaction-list-item
+          <!-- <transaction-list-item
```
saminakh, line 25: Why are pending transactions commented out?

staylorwr: Because we don't create or use them right now.  Recent SDK changes / rewrite mean we need to bring them back w/ the new events.

saminakh (approved):
Thank you for these fixes scott! The pagination is behaving better imo, and the TX history is now correctly unsubbing 🎉

PR #442, opened 2022-02-07
feat: Add disclaimer to wallet recovery

SK-Sam:
This PR adds an interstitial screen to remind users to re-add multiple accounts after recovery.

- Change of logic implemented so end of flow isn't on PIN confirmation
- New View added, wallet creation set to `I Understand` button now

### DIFF
--- a/src/views/CreateWallet/MultipleAccountsDisclaimer.vue
+++ b/src/views/CreateWallet/MultipleAccountsDisclaimer.vue
@@ -1,13 +1,16 @@
 <template>
-  <div>
-    <svg width="50" height="50" viewBox="0 0 50 50" fill="none" xmlns="http://www.w3.org/2000/svg">
-    <circle cx="25" cy="25" r="24" stroke="#052CC0" stroke-width="1.5"/>
-    <path d="M25 14V36" stroke="#052CC0" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
-    <path d="M14 25H36" stroke="#052CC0" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
-    </svg>
-  </div>
+    <div class="bg-rGrayLight text-center pt-8 pb-7">
+      <div class="mb-4 flex flex-row justify-center">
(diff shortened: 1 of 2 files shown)


SK-Sam:
@saminakh There was talks in the previous attempt at this feature to leave out adding an extra step to the flow, IIRC you suggested adding an extra step. What the logic to me is looking like is that `pinIsSet` as a bool is essentially acting like an extra step. Let me know if you still feel strongly about adding an extra step and I may just convert it towards that approach!

saminakh (approved):
This LGTM! The only thing i noticed was that testing locally, when we delete and restore, we aren't presented with the "accept TOS" screen until we attempt to change gateways. I don't know if they should 1) see this screen immediately, rather than after trying to change the gateway after deleting + restoring, or 2) shouldn't see the screen at all after deleting + restoring. This isn't related to your PR necessarily, i just happened to encounter it while testing out this PR...but we should maybe make a small issue for this/bring it up at sprint planning on wednesday maybe?

Otherwise the code LGTM and is working 👍

SK-Sam:
> This LGTM! The only thing i noticed was that testing locally, when we delete and restore, we aren't presented with the "accept TOS" screen until we attempt to change gateways. I don't know if they should 1) see this screen immediately, rather than after trying to change the gateway after deleting + restoring, or 2) shouldn't see the screen at all after deleting + restoring. This isn't related to your PR necessarily, i just happened to encounter it while testing out this PR...but we should maybe make a small issue for this/bring it up at sprint planning on wednesday maybe?
> 
> Otherwise the code LGTM and is working 👍

Thanks for the input for this and testing it thoroughly 😄 Totally worth making a note on this.

alexandreaco (approved):
I tested this locally and it works great. The code is clean and straightforward. Thanks Sam! 

One small nit that you can address if you have time:
When you have vue components with multiple attributes there's a convention to break those attributes up into multiple lines. See the style guide here https://v2.vuejs.org/v2/style-guide/?redirect=true#Multi-attribute-elements-strongly-recommended We haven't done a perfect job of implementing this across the app, but it's always a good time to start :)

PR #443, opened 2022-02-08
chore: add ledger decrypt message modal

saminakh:
This PR updates the wallet history page to include a modal when:
1) the current active account is a hardware account, and
2) when the wallet is waiting on the hardware ledger to confirm decryption

See loom demo:
https://www.loom.com/share/526c0941717d46adb1c7df48948f132e

### DIFF
--- a/src/views/Wallet/WalletHistory.vue
+++ b/src/views/Wallet/WalletHistory.vue
@@ -1,4 +1,5 @@
 <template>
+  <wallet-ledger-verify-decrypt-modal v-if="shouldShowDecryptModal" />
   <div class="flex flex-col flex-1 min-w-0 overflow-y-auto bg-white">
     <div class="bg-rGrayLightest py-6 px-8 bg-gray">
       <div class="flex justify-between">
@@ -88,18 +89,20 @@
 </template>
 
 <script lang="ts">
-import { computed, defineComponent, onMounted, watch } from 'vue'
+import { computed, ComputedRef, defineComponent, onMounted, watch } from 'vue'
 import TransactionListItem from '@/components/TransactionListItem.vue'
 import LoadingIcon from '@/components/LoadingIcon.vue'
 import ClickToCopy from '@/components/ClickToCopy.vue'
 import { useNativeToken, useWallet, useHistory } from '@/composables'
 import { useRouter, onBeforeRouteLeave } from 'vue-router'
+import WalletLedgerVerifyDecryptModal from './WalletLedgerVerifyDecryptModal.vue'
 
 const WalletHistory = defineComponent({
   components: {
     LoadingIcon,
     ClickToCopy,
-    TransactionListItem
+    TransactionListItem,
+    WalletLedgerVerifyDecryptModal
   },
 
   setup () {
@@ -108,6 +111,7 @@ const WalletHistory = defineComponent({
       activeAddress,
       activeAccount,
       explorerUrlBase,
+      hardwareAccount,
       radix,
       verifyHardwareWalletAddress
     } = useWallet(router)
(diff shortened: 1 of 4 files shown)


src/views/Wallet/WalletHistory.vue
```diff
@@ -141,8 +146,20 @@ const WalletHistory = defineComponent({
...
     })
 
+    const isHardwareAccount: Ref<boolean> = ref(false)
+    const shouldShowDecryptModal: ComputedRef<boolean> = computed(() => {
+      return isHardwareAccount.value && isDecrypting.value
+    })
+
     // Fetch new history when active account changes
-    watch((activeAccount), () => { resetHistory() })
+    watch((activeAccount), () => {
```
staylorwr, line 155: You shouldn't need two `watch` statements here, both are setting the `isHardwareAccount` value.

staylorwr (changes requested):

src/views/Wallet/WalletHistory.vue
saminakh: Doh! This was supposed to be watching the `hardwareAccount` value 🤦 I lost this logic when rebasing on top of main because i had forgotten to pull it down initially. Thank you for the catch!

saminakh: oh nevermind, i do have one watching `activeAccount` and one watching `hardwareAccount`. I'm tracking both here because when this component is first setup (when `useHistory` is first called) `hardwareAccount` is sometimes null. Is there a different way I can check to see if the user's currently viewing the history of a hardware account without watching both of these values?

src/views/Wallet/WalletHistory.vue
```diff
@@ -141,8 +146,20 @@ const WalletHistory = defineComponent({
       })
     })
 
+    const isHardwareAccount: Ref<boolean> = ref(false)
+    const shouldShowDecryptModal: ComputedRef<boolean> = computed(() => {
```
alexandreaco, line 150: minor nit: you can make this a 1 liner with an implicit return `const shouldShowDecryptModal: ComputedRef<boolean> = computed(() => isHardwareAccount.value && isDecrypting.value)`

src/views/Wallet/WalletHistory.vue
alexandreaco: I'm looking at this more and it seems like you're trying to do two different things.

1. `isHardwareAccount` is true when the active account is the hardware account. This can just be a computed property.
2. you reset history when the active account changes. This is a side effect and should live in a watch()

saminakh: @alexandreaco @staylorwr thank you for the help here! This should be cleaned up now! watching `hardwareAccount` did not appear to be necessary after all :D

src/views/Wallet/WalletHistory.vue
saminakh: i took it a step further and just put the logic for computing `isHardwareAccount` right inside `shouldShowDecryptModal`!

saminakh:
@alexandreaco @staylorwr could you look at this one last time? I cleaned it up after alex walked me through some changes after sprint planning!

src/views/Wallet/WalletHistory.vue
```diff
@@ -141,6 +146,11 @@ const WalletHistory = defineComponent({
       })
     })
 
+    const shouldShowDecryptModal: ComputedRef<boolean> = computed(() =>
```
alexandreaco, line 149: 👍

alexandreaco (approved):
Thanks for the updates!

staylorwr (approved):

PR #447, opened 2022-02-11
fix: have all token symbols clickable and redirects to Explorer token page

SK-Sam:
[Demo](https://www.loom.com/share/5cae67112bad4a1ea17d073a1411919a)

This PR adds the feature of clicking on any token symbol Olympia-wide and being directed to respective token's Explorer page. We are also exposing the token RRI by hovering over the token symbol.

We've adjusted `TokenSymbol.vue` to take required attributes `rri`, `symbol`, and `hasGreyBackground`. We initially wanted to pass the entire `Token`, but it seems that we're sometimes passing an `Action`. It may be better to pass these two strings `symbol` and `rri`, which both entities own as opposed to having the component figure out which of the Types it's extracting data from.

### DIFF
--- a/src/components/StakeListItem.vue
+++ b/src/components/StakeListItem.vue
@@ -64,8 +64,11 @@
             <div class="mb-1 w-26 flex-grow-0 text-rGrayMed text-xs">{{ $t('staking.pendingStakeLabel') }}:</div>
             <div class="mb-1 flex-1 text-rBlack">
               <big-amount :amount="pendingStakeAmount" />
-              <a :href="nativeTokenRRIUrl" target="_blank">
+              <a :href="nativeTokenRRIUrl" target="_blank" class="group cursor-pointer relative">
                 <span class="text-rGrayDark ml-1 uppercase">{{ nativeToken.symbol }}</span>
+                <div class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 rounded-sm shadow border border-solid border-rGrayLight">
+                  <span>{{ nativeToken && nativeToken.rri.toString() }}</span>
+                </div>
               </a>
             </div>
           </div>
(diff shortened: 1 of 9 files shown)


alexandreaco (commented):
Interesting use of slots here! Let's meet today to pair on this. I see a couple of places where we can simplify the TokenSymbol component. 

1. We are computing the explorer url in a few different places. If we instead pass in the whole Token type into TokenSymbol, we can encapsulate that logic in a single place
2. There are some places in the app that need this UX but don't need the grey box. Let's talk about how to add conditional css to make this work

SK-Sam:
> Interesting use of slots here! Let's meet today to pair on this. I see a couple of places where we can simplify the TokenSymbol component.
> 
>     1. We are computing the explorer url in a few different places. If we instead pass in the whole Token type into TokenSymbol, we can encapsulate that logic in a single place
> 
>     2. There are some places in the app that need this UX but don't need the grey box. Let's talk about how to add conditional css to make this work

Hey sure thing! Thanks a lot for the feedback, I agree with all of t.

I was actually treating the Explorer side as a fixup/refactor to these...less optimal approaches I did here! Only thing left would be what exactly to pass into `TokenSymbol.vue` in terms of a token, do we want the entire token? Or just the string?

I've made a re-usable component `TokenSymbol` in Explorer using class conditionals 😄 Let me know what you think of something like this, it seems to be working just fine but I'd love to hear what you think of the structure

```<template>
  <a
    class="group relative inline-flex font-light text-rGrayMark py-0.5 px-1 rounded borderself-end uppercase"
    :class="hasGreyBackground && 'bg-rGrayLight border border-rGray'"
    :href="tokenUrl"
  >
    {{ symbol }}
    <div class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 rounded-sm shadow border border-solid border-rGrayLight">
      {{ rriString }}
    </div>
  </a>
</template>```

File: src/views/Wallet/WalletOverview.vue [hunk 1]
```diff
@@ -215,6 +233,14 @@ const WalletOverview = defineComponent({
       })
     })
 
+    const nativeTokenRRIUrl: ComputedRef<string> = computed(() => {
```
alexandreaco, line 236: We no longer need this block now that we're using TokenSymbol

src/components/StakeListItem.vue
```diff
@@ -62,15 +62,39 @@
         <dl class="mt-1">
           <div v-if="validateGreaterThanZero(pendingStakeAmount)" class="flex items-center flex-wrap">
             <div class="mb-1 w-26 flex-grow-0 text-rGrayMed text-xs">{{ $t('staking.pendingStakeLabel') }}:</div>
-            <div class="mb-1 flex-1 text-rBlack"><big-amount :amount="pendingStakeAmount" /> <span class="text-rGrayDark ml-1 uppercase">{{ nativeToken.symbol }}</span></div>
+            <div class="mb-1 flex-1 text-rBlack">
```
alexandreaco, line 65: Can we update these three changes to use TokenSymbol as well?

alexandreaco:
Screen Shot 2022-02-15 at 11 52 36 AM
Screen Shot 2022-02-15 at 11 53 25 AM
@SK-Sam I spotted one bug where the TokenSymbol isn't rendering in the list of hidden balances. I think this is the last piece :)

SK-Sam:
Screen Shot 2022-02-15 at 12 01 33 PM

Thanks for this catch @alexandreaco , i see how I missed it. Didn't realize there were two syntaxes for calling the same component! d'oh.

alexandreaco:
@SK-Sam Please change the PR name to fix: so this doesn't cause a feature bump :)

alexandreaco (approved):

PR #448, opened 2022-02-11
fix: Load proper account balances upon initial render of wallet balances

SK-Sam:
[Demo](https://www.loom.com/share/425ea46981434bf8a866fc2068387403)

This PR handles which account's balances are being rendered upon completion of logging in. `useTokenBalances` was not handling latest account logic. The approach here is to create an `activeAccount` sub, or `activeAccountSub` which should be watched for any changes. This value will be passed into `useTokenBalances` when views are sensitive to latest account information that needs to be displayed, such as `WalletOverview.vue`.

### DIFF
--- a/src/components/HiddenTokenBalanceListItem.vue
+++ b/src/components/HiddenTokenBalanceListItem.vue
@@ -48,10 +48,10 @@ export default defineComponent({
 
   setup (props) {
     const router = useRouter()
-    const { radix } = useWallet(router)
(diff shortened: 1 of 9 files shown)


alexandreaco (approved):
:shipit:

saminakh (approved):
Code LGTM, and it's behaving well locally as far as I can tell! 👍

alexandreaco:

PR #450, opened 2022-02-15
fix: Don't load account 1 balances when first opening wallet

staylorwr:
This PR includes a collection of findings from @SK-Sam @alexandreaco and myself.  

The original UI was not handling changes to the `activeAccount` vue ref in the `useWallet` composable on the wallet overview.  This meant that when the `initWallet` function was called and it switched accounts there was a race condition that resulted in conflicting `activeAccount` values. _I think_.

This PR fixes these problems by updating the `WalletOverview` to delay displaying values until half a second after the balances observable has emitted an event.  This gives the UI enough time to correctly calculate and render the real balance values while the wallet switches and persists account values to the store and does... other things.

### DIFF
--- a/src/views/Wallet/WalletOverview.vue
+++ b/src/views/Wallet/WalletOverview.vue
@@ -82,7 +82,7 @@
 <script lang="ts">
 import { defineComponent, computed, ComputedRef, ref, Ref, onMounted, onUnmounted } from 'vue'
 import { merge, forkJoin, interval, Subject, Subscription } from 'rxjs'
-import { switchMap, mergeMap } from 'rxjs/operators'
+import { delay, switchMap, mergeMap } from 'rxjs/operators'
 import { Amount, AmountT, Token } from '@radixdlt/application'
 import BigAmount from '@/components/BigAmount.vue'
 import TokenSymbol from '@/components/TokenSymbol.vue'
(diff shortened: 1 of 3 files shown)


staylorwr:
fwiw I don't think the changes to `useTokenBalances` are necessary but I like the cleaner type references. 👍🏼

src/views/Wallet/WalletOverview.vue
```diff
@@ -150,19 +151,21 @@ const WalletOverview = defineComponent({
...
+
+    const balanceSub = balanceObs.subscribe(([stakes, unstakes]) => {
       activeStakes.value = stakes
       activeUnstakes.value = unstakes
-      loading.value = false
     })
-    subs.add(balanceSub)
 
+    subs.add(balanceSub)
+    subs.add(balanceObs.pipe(delay(500)).subscribe(() => { loading.value = false }))
```
SK-Sam, line 168: ⭐

SK-Sam (commented):
So far this PR has handled every issue i've noted being:
- Balance not updating after any form of tx
- Forms not loading due to `tokenBalance` being `null`

On top of this, it seems the account's are being chosen properly along with the respective balance upon logging in!

alexandreaco:

PR #451, opened 2022-02-16
fix: Watch active address instead of active account

alexandreaco:
Now that we know radix.activeAddress isn't updating fast enough, I thought I'd try relying on radix.activeAccount instead. It looks like it fixes the UI bug. However, if you switch around accounts with the balances view open on this branch and on main, the wallet makes exponentially more calls to the balances endpoint each time!


alexandreaco:

PR #452, opened 2022-02-16
fix: Update disconnect hardware device error shape for new component

alexandreaco:
This PR updates the error shape that's returned when disconnecting a hardware ledger to include the category attribute that was added in this other PR https://github.com/radixdlt/olympia-wallet/pull/438/files

### DIFF
--- a/src/composables/useTransactions.ts
+++ b/src/composables/useTransactions.ts
@@ -165,7 +165,13 @@ export default function useTransactions (radix: ReturnType<typeof Radix.create>,
     // To Do: Nicely handle hardware device error when full error is returned from API
     // For now: pattern match on transaction errors that return a "No device found" string or "DisconnectedDevice"
     if (err.toString().indexOf('Error: No device found') >= 0 || err.toString().indexOf('DisconnectedDevice') >= 0) {
-      setError({ ...err, type: 'HARDWARE' })
+      setError({
+        ...err,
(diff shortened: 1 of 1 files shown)


src/composables/useTransactions.ts
```diff
@@ -165,7 +165,14 @@ export default function useTransactions (radix: ReturnType<typeof Radix.create>,
     // To Do: Nicely handle hardware device error when full error is returned from API
     // For now: pattern match on transaction errors that return a "No device found" string or "DisconnectedDevice"
     if (err.toString().indexOf('Error: No device found') >= 0 || err.toString().indexOf('DisconnectedDevice') >= 0) {
-      setError({ ...err, type: 'HARDWARE' })
+      // setError({ ...err, type: 'HARDWARE' })
```
SK-Sam, line 168: Maybe a leftover comment!

SK-Sam (approved):
This is great. the hardware disconnect flow is working for me just fine! left a note for what could be a leftover comment

saminakh (approved):
This seems reasonable to me! Could you explain to me what was breaking previously though? looking at `ErrorModalHardware.vue`, i feel like the `if (type)` should have been sufficient (wouldn't `type` be undefined, so the default values of the message and title are used and the switch statement is skipped over?) I'm not sure what i'm missing 😬

PR #453, opened 2022-02-16
chore: Present users with a better error experience if they loose network connectivity

alexandreaco:
Screen Shot 2022-02-16 at 2 05 29 PM

While we wait for better network connectivity error messages from the SDK, we are pinging the network every 10 seconds and showing users a warning if the network cannot be reached. In this PR, we are removing the cancel button and added some warning text encouraging users not to re-attempt transactions that may have succeeded before connection was lost. Removing the cancel button will prevent users from continuing to use the app when it's not connected to the network.


SK-Sam (approved):
Cool! Just saw the convo where the warning text was approved 👍

PR #455, opened 2022-02-17
fix: Proper precision for user-entered amounts.

Mleekko:
To reproduce the bug, try to enter `11.111111111111111111` when sending the transaction.
This gets converted to `11111111111111110000` attos, while should be sent as "11111111111111111111".
Looks like this fixes the leftover dust when unstaking too.

### DIFF
--- a/src/helpers/validateRadixTypes.ts
+++ b/src/helpers/validateRadixTypes.ts
@@ -19,7 +19,7 @@ export const safelyUnwrapValidator = (validatorString: string): ValidatorAddress
   return validatorAddressRes.value
 }
 
-export const safelyUnwrapAmount = (amount: number): AmountT | null => {
+export const safelyUnwrapAmount = (amount: string): AmountT | null => {
   const bigAmount = new BigNumber(amount)
   const amountInput = bigAmount.shiftedBy(18) // Atto
   const amountResult = Amount.fromUnsafe(amountInput.toFixed())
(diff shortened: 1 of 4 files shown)


Mleekko:
@SK-Sam @alexandreaco 
Is there any way to get this merged?

SK-Sam:
Hey @Mleekko , thanks for bringing up this PR! Currently we have other structural changes that may affect the QA process of your PR. This is on our list to work with, but we'd like to get other changes out the door first! Just wanted to let you know this PR has visibility, and we can ping you as soon as we get to this!

Thanks so much again @Mleekko

Mleekko:
Is there any chance that this gets into the next release?

artsycoder533 (approved):
I completed two identical `send token` transactions in both the current build of the app and the build from this branch.  In the current build of the app, if the user amount entered is `11.111111111111111111` when you send tokens, it shows in the history as only `11.11111111111111` being sent, a loss of 4 significant digits.  In the build associated with the pr, I completed the same transaction and in the history the amount is shown as `11.111111111111111111` which matches the same amount of digits initially entered by the user.

It’s smart to change the amount to be of type string.  According to `bignumber.js` documentation, the BigNumber constructor accepts a string, number, or BigNumber as a parameter.  The benefit to passing in a string to the constructor is, according to the docs, there is no limit to the number of digits of a value of type string.  Instead of parsing the string to a number and being limited to the number of digits javascript can handle, the entire amount the user entered  is passed into `safelyUnwrapAmount`.  This prevents precision loss up to the limitations of the package.

In the demo video, the transaction completed from this PRs build is labeled as `testing pr 455`.  The transaction completed from the current build of the app is labeled as `current build`. On hover you can see the difference in the amounts of each transaction.


https://user-images.githubusercontent.com/83678228/173598085-066e2321-6faa-4173-a611-6d9bac090c5a.mov

PR #454, opened 2022-02-17
Do not merge: brainstorming wallet history decrypt bug

saminakh:
@SK-Sam check this out!

PR #456, opened 2022-02-17
fix: decrypt errors while switching accounts on history page

SK-Sam:
[Demo showcasing freely switching between accounts on History Page](https://www.loom.com/share/8dcd1a5547aa491f8ead3949c45e4236)

Joint effort by @saminakh & I! https://github.com/radixdlt/olympia-wallet/pull/454/files

This PR updates account switching logic in `useHistory` composeable, and updates the `account` value before resetting History Page transactions while switching accounts. We're also making sure to restart polling of subscriptions properly upon account changing.

The issue previously was that the incorrect `publicKey` was being used from the incorrect account, while it seemed like the `to_address` public key and `from_address` public key was correct.

We're also now handling hardware-account-specific account changing. Previously, certain modals such as decrypt on hardware device modal would not fire off on History page.

### DIFF
--- a/src/composables/useHistory.ts
+++ b/src/composables/useHistory.ts
@@ -33,7 +33,7 @@ export default function useHistory (radix: ReturnType<typeof Radix.create>, acco
     }
   }
 
-  const cleanupHistorySub = () => {
+  const cleanupTransactionSub = () => {
     if (transactionSub) {
       transactionSub.unsubscribe()
       transactionSub = null
@@ -41,7 +41,7 @@ export default function useHistory (radix: ReturnType<typeof Radix.create>, acco
   }
 
   const resetHistory = () => {
-    cleanupHistorySub()
+    cleanupTransactionSub()
     loadingHistory.value = true
     cursorStack.value = []
     transactions.value = []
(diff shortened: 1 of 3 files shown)


src/composables/useHistory.ts
```diff
@@ -16,8 +16,9 @@ const transactions: Ref<SimpleExecutedTransaction[]> = ref([])
 
 const isDecrypting: Ref<boolean> = ref(false)
 
-export default function useHistory (radix: ReturnType<typeof Radix.create>, activeAccount: AccountT) {
+export default function useHistory (radix: ReturnType<typeof Radix.create>, account: AccountT) {
   let transactionSub: Subscription | null
+  let activeAccount: AccountT = account
```
alexandreaco, line 21: This looks like it's on the right path. However, we need to make this `activeAccount` a ref to guarantee that the watcher below will work correctly

`const activeAccount: Ref<AccountT> = ref(account)`

SK-Sam: great, thanks for explaining this concept!

SK-Sam:
@alexandreaco i've found a hole in logic which may cause certain modals not to pop up. One case is in History Page, where if you switch accounts into a Hardware Account, you won't see HW-based modals after the second switch

alexandreaco (approved):
We paired on this and I saw it working. Thank you @SK-Sam!

src/composables/useHistory.ts
```diff
@@ -31,14 +33,25 @@ export default function useHistory (radix: ReturnType<typeof Radix.create>, acti
     }
   }
 
+  const cleanupHistorySub = () => {
```
caleboller, line 36: Very minor nit, seems like we could clean up naming here a bit? I'm not sure which is more accurate, `historySub` or `transactionSub`, but can we settle on one?

caleboller (approved):

PR #457, opened 2022-02-17
chore: Don’t show validator uptime for validators below top 100

alexandreaco:
Screen Shot 2022-02-17 at 8 12 03 AM

We already had this feature in the explorer. Now, we will also hide recent uptime for validators not in the top 100 for the wallet.

### DIFF
--- a/src/components/StakeListItem.vue
+++ b/src/components/StakeListItem.vue
@@ -55,7 +55,7 @@
               class="mb-1 flex-1 text-rBlack"
               :class="{'text-rRed': Number(validator.uptimePercentage) <= 98}"
             >
-              {{validator.uptimePercentage}}%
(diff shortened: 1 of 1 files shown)


src/components/StakeListItem.vue
```diff
@@ -186,6 +186,10 @@ const StakeListItem = defineComponent({
       return checkValidatorUrlExploitable(validator.value.infoURL.toString())
     })
 
+    const validatorUptimeContent: ComputedRef<string> = computed(() => {
+      return inTopOneHundred.value && validator.value ? `${validator.value.uptimePercentage}%` : '-'
```
SK-Sam, line 190: 💯

SK-Sam (approved):
LGTM & locally spoofed just to make sure

PR #458, opened 2022-02-17
fix: Use latest SDK function to verify password

alexandreaco:
Previously, we were still using an old SDK function to verify the user's password before resetting it. This was always resolving to true, even when the user put in an incorrect password. This PR updates the flow to use the same function we use when logging in the user. Now, you must put in a correct current password to update your password in settings.

### DIFF
--- a/src/views/Settings/SettingsResetPassword.vue
+++ b/src/views/Settings/SettingsResetPassword.vue
@@ -52,8 +52,7 @@
 <script lang="ts">
 import { computed, ComputedRef, defineComponent, onUnmounted, Ref } from 'vue'
 import { useForm } from 'vee-validate'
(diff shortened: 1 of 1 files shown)


src/views/Settings/SettingsResetPassword.vue
```diff
@@ -104,23 +103,22 @@ const SettingsResetPassword = defineComponent({
...
-        } else {
-          isLoading.value = false
-          setErrors({
-            currentPassword: t('validations.incorrectPassword')
-          })
-        }
-      } catch {
-        isLoading.value = false
+
+      const signingKeychainResult = await SigningKeychain.byLoadingAndDecryptingKeystore({
```
SK-Sam, line 109: 👍

PR #459, opened 2022-02-18
fix: Correctly load token balances when switching active account

alexandreaco:
This PR takes a fresh approach to solving the active account and subscription issues on the balances view. Thanks to @SK-Sam and @staylorwr for all of the brainstorming. The original bug described a flash on the balances screen with the wrong account balances appearing before the real active account balance appeared. It turns out, this was an issue on all views that show token balances and allow for account switching from the sidebar. 

This PR fixes the bug on the following pages. At the time of posting, we still need to fix the settings > hidden tokens page

- [x] Token Balances
- [x] Transfer Tokens
- [x] Stake/Unstake Tokens
- [x] Settings Hidden Tokens Thanks to @SK-Sam for help on this. I merged some of his ideas from https://github.com/radixdlt/olympia-wallet/pull/462/commits/beb023ecd137fe90f4bce5270aa2a308149de305

### DIFF
--- a/src/composables/useTokenBalances.ts
+++ b/src/composables/useTokenBalances.ts
@@ -22,16 +22,23 @@ export default function useTokenBalances (radix: ReturnType<typeof Radix.create>
 
     // Get token info for tokens related to this account and save in memory
     // Don't save duplicates
+    const relatedTokenPromises: Promise<any>[] = []
     res.account_balances.liquid_balances.map((balance) => {
       if (!relatedTokens.value.find((t) => t.rri.equals(balance.token_identifier.rri))) {
-        firstValueFrom(radix.ledger.tokenInfo(balance.token_identifier.rri))
-          .then((t) => {
-            if (!relatedTokens.value.find((t) => t.rri.equals(balance.token_identifier.rri))) {
-              relatedTokens.value = [...relatedTokens.value, t]
-            }
-          })
+        relatedTokenPromises.push(
+          firstValueFrom(radix.ledger.tokenInfo(balance.token_identifier.rri))
+            .then((t) => {
+              if (!relatedTokens.value.find((t) => t.rri.equals(balance.token_identifier.rri))) {
+                relatedTokens.value = [...relatedTokens.value, t]
+              }
+            })
+        )
       }
     })
+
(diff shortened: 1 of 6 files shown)


staylorwr:
https://www.loom.com/share/3d2b142dcc7c49d29f4b618592e5bc9b

src/composables/useTokenBalances.ts
```diff
@@ -1,33 +1,46 @@
...
-      })
+        )
+      }
     })
 
+    // Wait until missing related tokens have been fetched before returning so all token info is present
+    // for account balances
+    await Promise.all(relatedTokenPromises)
+  }
+
```
SK-Sam, line 43: This must be the complete fix for hidden balances! I was running into an issue where token balances were always returning from a previous account only once right after switching to a new account, and every switch after proceeded w/ proper token balances

👍

src/views/Settings/SettingsTokens.vue
```diff
@@ -75,26 +74,52 @@ export default defineComponent({
...
+    /* ------
+     *  Functions
+     */
+    const computeListOfHiddenTokens = (newActiveAddress: AccountAddressT) => {
+      loading.value = true
+      fetchBalancesForAddress(newActiveAddress)
+        .then(getHiddenTokens)
+        .then((res) => {
+          // Update hidden tokens when active address change
+          const hiddenTokensForActiveAccount = res.filter((token) => tokenBalanceForByString(token) !== null)
```
SK-Sam, line 107: 👍

SK-Sam (approved):
This is great! Looks like the best approach out of our three PR's w/ account switching logic. I gave Hidden Tokens issue a spin and all seems to work well 😄

src/composables/useTokenBalances.ts
alexandreaco: yeah, promises and race conditions will get you every time. I tested this by setting an account with only a few hidden tokens as my "latest account." Then, rebooting the app, going to the hidden balances screen, and then switching to an account with many hidden tokens (including the ones from the first account). Now the view waits until all of the little token symbol badges can render properly

staylorwr (approved):
👍🏼 to handling validation errors later

PR #462, opened 2022-02-23
chore: filter hidden tokens against current wallet

SK-Sam:
We're now going to filter the current wallet balance's tokens on SettingsToken page based on which token RRI's are hidden on the active account. I'm running into one issue where on first render, changing to another account which had different hidden tokens than initial account MAY take an additional account switch. Currently investigating but I believe this solves majority of the issue!

### DIFF
--- a/src/views/Settings/SettingsTokens.vue
+++ b/src/views/Settings/SettingsTokens.vue
@@ -7,10 +7,10 @@
       <div class="text-rGrayDark text-sm max-w-lg mb-8">{{ $t('settings.tokensDescription') }}</div>
 
       <HiddenTokenBalanceListItem
-        v-for="token in hiddenTokens"
-        :key="token"
-        :tokenRRI="token"
-        :tokenBalance="tokenBalanceForByString(token)"
+        v-for="token in tokenBalanceOfHidden"
+        :key="token.token_identifier.rri.toString()"
+        :tokenRRI="token.token_identifier.rri.toString()"
+        :tokenBalance="tokenBalanceForByString(token.token_identifier.rri.toString())"
         @select="handleRequestUnhideToken"
       />
     </template>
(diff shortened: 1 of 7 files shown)


src/views/Settings/SettingsTokens.vue
```diff
@@ -89,18 +89,9 @@ export default defineComponent({
     })
 
     watch((activeAddress), (newActiveAddress) => {
-      // Update balances when active address changes
       newActiveAddress && fetchBalancesForAddress(newActiveAddress)
     })
 
-    watch((tokenBalances), () => {
-      getHiddenTokens().then((res) => {
```
saminakh: Why was this code removed? Is it because we fetch hidden tokens on mount, and since you have to leave this page to hide a token you don't have to worry about tokens being hidden while just passively looking at the page?

saminakh (approved):
This code LGTM, and appears to be working well locally. I do think it's interesting that we are filtering out hidden tokens for those that have a 0 balance in a particular account. I almost feel like the original tab name ("Balances") was more appropriate than the current one ("Tokens") with that filtering in place, whereas if there was no filtering maybe "Tokens" would be more applicable. But I don't think that one's up for us to decide 😛

saminakh:
Side note, but is there a linear issue for this? I was looking at linear to see the exact requirements and couldn't find one 🤔

src/views/Settings/SettingsTokens.vue
SK-Sam: Yes! That is it. I believe watching `activeAddress` change and fetching for a balance was more important here. @alexandreaco touched up a race condition where with this `watch((tokenBalances)...` snippet I had, it would be a fraction of a second too late to actually update the token balance, even while being watched 😦 

In `useTokenBalances` @alexandreaco has utilized a promise along with waiting for the token balance to update, THEN to continue on. I tried to achieve this with the `watch` you highlighted here, but didn't address the race condition properly.

SK-Sam:
> Side note, but is there a linear issue for this? I was looking at linear to see the exact requirements and couldn't find one 🤔

Sadly no Linear issue for this 🐸 I believe this surfaced with doing a full application QA upon Alex's composable refactor! You make a good point, it should be there 🤔

alexandreaco:

PR #464, opened 2022-02-25
fix: transaction form prematurely showing amount error 

P-v-R:
On the transactions page, the amount form field prematurely showed an error message on token refresh caused by a bug existing in an older version of vee-validate / rules. Updating both packages from `4.4.5` to `4.5.9` fixed the issue.


alexandreaco (approved):
Hey this fix works, thanks! I'm glad it was a simple fix :)

PR #466, opened 2022-02-28
fix: staking amount error prematurely displays

P-v-R:
See [original error ](https://www.loom.com/share/3d2b142dcc7c49d29f4b618592e5bc9b) here.

When in the staking view, if a user chose to stake / unstake without preemptively filling out the amount form field, a `valid amount` error would occur, and would not disappear until the user clicked out of the form component (IE on `blur`). The chances of a user filling out the amount form field before choosing whether to stake or unstake were unlikely so the error no longer appears. 

This was caused by the stake/unstake buttons triggering a validation function that checks the values of both the automatically populated `validator address` and the `amount` value. The validation assumes that a value will be passed for `values.amount` and raises an error when the amount is left undefined. Now if the `values.amount === undefined` we bypass the validation and then `validateOnInput` of the form field.   

radix-slow

### DIFF
--- a/src/views/Wallet/WalletStaking.vue
+++ b/src/views/Wallet/WalletStaking.vue
@@ -257,16 +257,14 @@ const WalletStaking = defineComponent({
       setActiveForm('STAKING')
       setActiveTransactionForm('stake')
       values.validator = validator.toString()
-      if (values.amount === undefined) return
-      validate()
+      if (values.amount) validate()
     }
 
     const handleReduceFromValidator = (validator: ValidatorAddressT) => {
       setActiveForm('UNSTAKING')
       setActiveTransactionForm('unstake')
       values.validator = validator.toString()
-      if (values.amount === undefined) return
-      validate()
+      if (values.amount) validate()
     }
 
     const handleSubmitStake = () => {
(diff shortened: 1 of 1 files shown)


src/views/Wallet/WalletStaking.vue
```diff
@@ -256,13 +257,15 @@ const WalletStaking = defineComponent({
       setActiveForm('STAKING')
       setActiveTransactionForm('stake')
       values.validator = validator.toString()
+      if (values.amount === undefined) return
```
alexandreaco, line 260: I tried this out and it works, nice fix! One thing to note: a more "javascript-y" way of writing this logic would be to listen for the truthy case, do something, and let the falsy case be the default. 

For example
```js
if (values.amount) validate()
```

or 
```js
values.amount && validate()
```

P-v-R: Good note @alexandreaco ! I opted for the first suggestion, I see it to be a bit more readable. Just pushed the change.

alexandreaco (approved):
💯 Thanks @P-v-R 🙇‍♀️

PR #468, opened 2022-03-02
fix: staking prematurely displays error

P-v-R:
[Extension of this pull request ](https://github.com/radixdlt/olympia-wallet/pull/466)

We found an additional premature form field error popping up when a user clicks into either 'address' or 'amount', clicking out of the form field component entirely causes an error message to appear even when `add/reduce stake` is clicked. An easy way around this was to disable validation on blur - and as an extra level of protection, add `validateOnInput = true` to the address field.

📸 - *notice at end I delete a character from the address and the error message appears*
radixValidFix

### DIFF
--- a/src/views/Wallet/WalletStaking.vue
+++ b/src/views/Wallet/WalletStaking.vue
@@ -51,7 +51,6 @@
               :placeholder="$t('staking.validatorPlaceholder')"
               rules="required|validValidator"
               :validateOnInput="true"
-              :validateOnBlur="false"
             />
             <FormErrorMessage name="validator" class="text-sm text-red-400" />
           </div>
@@ -67,7 +66,6 @@
                   :placeholder="amountPlaceholder"
                   rules="required|validAmount"
                   :validateOnInput="true"
-                  :validateOnBlur="false"
                 />
                 <FormErrorMessage name="amount" class="text-sm text-red-400" errorClass="w-120" />
               </div>
@@ -147,7 +145,7 @@ const WalletStaking = defineComponent({
   setup () {
     const router = useRouter()
     const { t } = useI18n({ useScope: 'global' })
-    const { errors, values, meta, setErrors, resetForm, validate } = useForm<StakeForm>()
+    const { errors, values, meta, setErrors, resetForm, validateField } = useForm<StakeForm>()
     const {
       activeAddress,
       activeAccount,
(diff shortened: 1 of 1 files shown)


P-v-R:
fixing merge conflict now

src/views/Wallet/WalletStaking.vue
```diff
@@ -50,6 +50,8 @@
               class="w-full text-sm font-mono placeholder-sans"
               :placeholder="$t('staking.validatorPlaceholder')"
               rules="required|validValidator"
+              :validateOnInput="true"
+              :validateOnBlur="false"
```
alexandreaco, line 54: Validations are so tricky! Unfortunately not validating on blur introduces more bugs. If we are not validating on blur, then the warnings don't appear. But the form is also not being validated. That means that the submit button will not be enabled unless the user goes back and manually inputs text into both fields. See this loom where I press "add stake", enter an amount, and attempt to submit the form https://www.loom.com/share/af3bcae11e7140bd9650d0c60ef89be1

This is a case where representing the problem in a [DDD style](https://martinfowler.com/bliki/GivenWhenThen.html) can help illuminate the problem. Here is the first part of the bug report.

GIVEN a user is presented with an empty staking form
WHEN a user clicks into an input box for a required field
AND they click out of the input box
THEN they are presented with a validation error

☝️ This is expected behavior because both the validator and amount are required fields!

Let's look at the rest of the bug report...

GIVEN a user is presented with a staking form with validation errors because of missing required fields
WHEN a user clicks "add stake" or "remove stake"
THEN they should see the validator ID field populated
AND they should see the amount field cleared
**AND they should see the validation warnings cleared**

☝️ here is where the bug occurs. How can we write a solution that clears the form's validation when the user clicks "add stake" or "remove stake"?

alexandreaco (changes requested):

src/views/Wallet/WalletStaking.vue
P-v-R: @alexandreaco thanks for clearing this up. Now the `add/reduce stake` buttons will clear the form fields & corresponding error messages, assign the staking validator address and validate it, and will monitor the amount field to validate on new input. [see video and brief explanation here ](https://www.loom.com/share/5b3c8dc79f7c4f9390e984878e1e76b2)

staylorwr (approved):

PR #469, opened 2022-03-03
chore: Add max button for unstaking

SK-Sam:
[Demo](https://www.loom.com/share/3aaa659c442e4e2c9efa2afdcd6d18d3)

This PR implements a new button which allows User to unstake 100% of their XRD from validators. We add extra logic in order to switch between "Max Unstake Mode" and regular unstake mode.

No unique step has to be taken to unstake via Max Mode, User simply hits the same `Request Unstake` button as I was able to implement a function that decides which unstake mode to use.

Screen Shot 2022-03-03 at 9 22 40 AM

Screen Shot 2022-03-03 at 9 22 55 AM

Screen Shot 2022-03-03 at 10 18 51 AM

### DIFF
--- a/src/components/BigAmount.vue
+++ b/src/components/BigAmount.vue
@@ -1,5 +1,8 @@
 <template>
-  <span class="relative inline-flex flex-col items-center group cursor-pointer" @click.stop="copyText">
+  <span v-if="showMaxUnstakeText">
+    <span>MAX</span>
+  </span>
+  <span v-else class="relative inline-flex flex-col items-center group cursor-pointer" @click.stop="copyText">
     <span>{{numberForDisplay}}</span>
     <div class="absolute invisible group-hover:visible -mt-full bg-rGrayLightest text-rBlack bottom-full text-xs p-1 left-0 rounded-sm shadow border border-solid border-rGrayLight">
       {{fullNumber}}
@@ -105,6 +108,11 @@ const BigAmount = defineComponent({
     amount: {
       type: Object as PropType<AmountT>,
       required: true
+    },
+    showMaxUnstakeText: {
+      type: Boolean,
+      required: false,
+      default: false
     }
   },
 
--- a/src/components/BigAmount.vue
+++ b/src/components/BigAmount.vue
@@ -1,6 +1,6 @@
 <template>
   <span v-if="showMaxUnstakeText">
-    <span>MAX</span>
+    {{ $t('staking.maxUnstakeCapitalized') }}
   </span>
   <span v-else class="relative inline-flex flex-col items-center group cursor-pointer" @click.stop="copyText">
     <span>{{numberForDisplay}}</span>
@@ -16,6 +16,7 @@ import { defineComponent, PropType } from 'vue'
 import { AmountT } from '@radixdlt/application'
 import BigNumber from 'bignumber.js'
 import { useToast } from 'vue-toastification'
+import { useI18n } from 'vue-i18n'
 
 BigNumber.set({
   ROUNDING_MODE: BigNumber.ROUND_HALF_UP,
@@ -118,6 +119,7 @@ const BigAmount = defineComponent({
 
   setup () {
     const toast = useToast()
+    const { t } = useI18n({ useScope: 'global' })
     return { toast }
   },
 
--- a/src/views/Wallet/WalletConfirmTransactionModal.vue
+++ b/src/views/Wallet/WalletConfirmTransactionModal.vue
@@ -254,10 +254,10 @@ const WalletConfirmTransactionModal = defineComponent({
       return ''
     })
 
-    const amount: ComputedRef<AmountOrUnsafeInput | undefined> = computed(() => {
+    const amount: ComputedRef<AmountOrUnsafeInput> = computed(() => {
       if (stakeInput.value) {
         return stakeInput.value.amount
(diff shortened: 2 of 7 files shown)


src/components/BigAmount.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <span class="relative inline-flex flex-col items-center group cursor-pointer" @click.stop="copyText">
+  <span v-if="showMaxUnstakeText">
+    <span>MAX</span>
```
alexandreaco, line 3: This string is in text.js, can you pull it in through i18n? You also don't need the second nested `<span>` tag here

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -53,7 +53,8 @@
           <div class="border-b border-rGray py-3.5 flex items-center">
             <div class="w-26 text-right text-rGrayDark mr-6">{{ $t('transaction.amountLabel') }}</div>
             <div class="flex-1 flex flex-row items-center">
-              <big-amount :amount="amount" class="mr-1" />
+              <big-amount v-if="!shouldShowMaxUnstakeConfirmation" :amount="amount" class="mr-1" />
```
alexandreaco, line 56: The conditional reads a bit strangely. What do you think about changing the order to watch for the truthy value and let the default case be the old behavior?
```js
<span
  v-if="shouldShowMaxUnstakeConfirmation"
  class="mr-1 uppercase"
>
    {{ $t('staking.maxUnstakeButton') }}
</span>
<big-amount
  v-else
  :amount="amount"
  class="mr-1"
/>
```

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -252,7 +254,7 @@ const WalletConfirmTransactionModal = defineComponent({
       return ''
     })
 
-    const amount: ComputedRef<AmountOrUnsafeInput> = computed(() => {
+    const amount: ComputedRef<AmountOrUnsafeInput | undefined> = computed(() => {
```
alexandreaco, line 257: Why can this be undefined now?

src/views/Wallet/WalletStaking.vue
```diff
@@ -297,16 +329,57 @@ const WalletStaking = defineComponent({
         })
     }
 
+    const handleMaxSubmitStake = () => {
+      if (!nativeToken.value) return
+      const safeAddress = safelyUnwrapValidator(values.validator)
+      if (!safeAddress) return
+      // const safeAmount = safelyUnwrapAmount(Number(0.0000000000000001))
```
alexandreaco, line 336: I think you can remove this line :)

src/views/Wallet/WalletStaking.vue
```diff
@@ -297,16 +329,57 @@ const WalletStaking = defineComponent({
...
+      const safeAmount = safelyUnwrapAmount(Number('0.0000000000000001'))
+      if (!safeAmount) return
+
+      unstakeTokens({
```
alexandreaco, line 340: Why does staking a single atto make the user unstake max?

src/views/Wallet/WalletConfirmTransactionModal.vue
SK-Sam: We now check for existence of `unstakeInput.value` and `unstakeInput.value.amount` to avoid `undefined`

src/views/Wallet/WalletStaking.vue
SK-Sam: i believe 1 atto === 100% when sending an amount to the Radix API. Shifting by 3 zero's to the left causes an error where a num must be > 0. Shifting by 1 zero to the right causes an error where a num must be < 100.

src/views/Wallet/WalletConfirmTransactionModal.vue
SK-Sam: Thanks for the heads up 😄 i'll try to avoid using `not`'s on my boolean vars to have them read better. This works fine!

src/views/Wallet/WalletStaking.vue
SK-Sam: renamed var to `safeOneHundredPercent`

src/views/Wallet/WalletConfirmTransactionModal.vue
```diff
@@ -255,7 +266,7 @@ const WalletConfirmTransactionModal = defineComponent({
     const amount: ComputedRef<AmountOrUnsafeInput> = computed(() => {
       if (stakeInput.value) {
         return stakeInput.value.amount
-      } else if (unstakeInput.value) {
+      } else if (unstakeInput.value && unstakeInput.value.amount) {
```
loganhenson, line 269: You could potentially use optional chaining here to make this a tiny bit tidier, e.g. `unstakeInput.value?,amount`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

SK-Sam: oh sweet this is much prettier 😄 thanks!

staylorwr (commented):
Looks great!

staylorwr (approved):

PR #472, opened 2022-03-16
fix: display error modal on failed hardware decryption

P-v-R:
When a user is logged in to a hardware account, and the ledger device is disconnected, unhandled console errors would appear when the user tried to decrypt an encrypted message on the transaction history page. Now when a user encounters this situation, an error modal will pop up and inform the user of an error that is most likely occurred from a disconnected ledger.

[Here is an example of the pop-up in action.](https://www.loom.com/share/f80e68f7b041486d9548a54121c9ddfc)

📸📸📸📸
Screen Shot 2022-03-16 at 8 30 38 AM

### DIFF
--- a/src/composables/useHistory.ts
+++ b/src/composables/useHistory.ts
@@ -6,6 +6,7 @@ import { formatWalletAddressForDisplay } from '@/helpers/formatter'
 const PAGE_SIZE = 30
 
 const decryptedMessages: Ref<{id: string, message: string}[]> = ref([])
+const displayLedgerErrorModal: Ref<boolean> = ref(false)
 const cursorStack: Ref<string[]> = ref([])
 const canGoBack: ComputedRef<boolean> = computed(() => cursorStack.value.length > 0)
 const canGoNext: Ref<boolean> = ref(false)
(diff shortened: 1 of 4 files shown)


src/composables/useHistory.ts
```diff
@@ -54,9 +55,18 @@ export default function useHistory (radix: ReturnType<typeof Radix.create>, acco
...
   const decryptMessage = (tx: ExecutedTransaction) => {
     isDecrypting.value = true
-    firstValueFrom(radix.decryptTransaction(tx)).then((val) => {
-      decryptedMessages.value.push({ id: tx.txID.toString(), message: val })
-    }).finally(() => { isDecrypting.value = false })
+    firstValueFrom(radix.decryptTransaction(tx))
+      .then((val) => {
+        decryptedMessages.value.push({ id: tx.txID.toString(), message: val })
+      })
+      .catch(() => {
```
saminakh, line 62: 💥

saminakh (approved):
This code LGTM and is working locally! 🎉 I think maybe we're starting to approach a point where we should reconcile some of the extra modals being made...we have a generic one, and a couple of specific ones for hardware stuff, maybe we should have a `genericHardwareModal` that can handle a couple of error types? but just a thought for the future! I think this is good to go 👍

PR #474, opened 2022-03-22
chore: create logic to refresh app after set time and requirements

SK-Sam:
[Demo](https://www.loom.com/share/b840b11d338d4c738161f592908c3883)

This PR counts mouse movement and key presses as interactions and resets application if no interactions were logged within one hour.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -4,6 +4,7 @@ import electron, { app, ipcMain, protocol, BrowserWindow, webContents } from 'el
 import { createProtocol } from 'vue-cli-plugin-electron-builder/lib'
 import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer'
 import path from 'path'
+import { debounce } from '@/helpers/debounce'
 import contextMenu from 'electron-context-menu'
 import { copyToClipboard, getKeystoreFile, storePin, validatePin, writeKeystoreFile } from '@/actions/electron/create-wallet'
 import {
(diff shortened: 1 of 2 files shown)


src/background.ts
```diff
@@ -113,6 +113,29 @@ app.on('ready', async () => {
...
+  win.webContents.on('cursor-changed', (_, type) => {
+    if(type) {
+      interactionCounter += 1
+    }
+  })
+
+  // Window refreshes if no interaction & 1hr passed
+  setInterval(() => {
+    if(interactionCounter === 0) {
+      win.reload()
```
alexandreaco, line 134: This is a great and simple solution 🙌

alexandreaco (approved):

src/background.ts
caleboller: @alexandreaco I have to disagree with your assessment here. I think there's room for improvement in this solution. 

If I'm parsing this logic correctly, this timeout operates in hour long intervals and then checks `interactionCounter` for activity in the last hour interval. What this most likely means is that our time to an actual "idle" point is quite likely longer than hour, up to 1:59:59.   For example, a user is in the middle of a 60 minute window, they stop interactions at the 30 minute mark, for that window they're considered active, it wouldn't be to the close of the next hour longer window, after 90 minutes of idle time, that they'd be recognized as idle.

I think a solution that meets the spec here would actually simplify the code a bit as well. What I've seen implemented in cases like this is a simple interval, like:

`let idleInterval = setInterval(() => { win.reload() }, 36000)`

And then for each event that would indicate an activity from a user, we'd clear that interval and recreate it, for example:

```
win.webContents.on('before-input-event', (_, input) => {
    if(input.type === 'keyDown') {
      // refactor this into its own function as we'll be calling it in multiple places
      clearInterval(idleInterval)
      idleInterval = setInterval(() => { win.reload() }, 36000)
    }
  })
```

Does that make sense?

caleboller (changes requested):

SK-Sam:
@caleboller i've refactored based on your suggestions and it does seem simpler and handles the issue much better :) thanks https://www.loom.com/share/b840b11d338d4c738161f592908c3883

src/background.ts
```diff
@@ -113,6 +118,23 @@ app.on('ready', async () => {
...
+  // interact with keyboard, refresh app and have User log in again.
+  let idleInterval = setInterval(() => { win.reload()}, 3600000)
+
+  win.webContents.on('before-input-event', (_, input) => {
+    if(input.type === 'keyDown') {
+      idleInterval = resetInteractionTimer(idleInterval)
+    }
+  })
+
+  win.webContents.on('cursor-changed', (_, type) => {
```
saminakh, line 132: Did we need to incorporate any debounce or throttling here? the interval will get reset a lot w/ cursor movements...are there any performance impacts that we should care about (that would be improved with the debounce/throttling)?

saminakh (commented):
This is working for me! I left one thought in case it's helpful at all!

staylorwr (approved):

caleboller (approved):

PR #475, opened 2022-03-23
fix: Max Unstaking button should only fire for when unstake form is active

alexandreaco:
https://www.loom.com/share/ff7cf1af26ca4d26a9a8980414710930

This PR fixes two bugs
1. Toggling between the stake and unstake forms should reset the "max unstake" status
2. Pressing stake should never unstake tokens


caleboller (approved):
Looks great! Thanks Alex.

PR #478, opened 2022-04-07
epic: Multiple Hardware Accounts (WIP)

staylorwr:
WIP PR for multiple hardware devices with multiple hardware accounts.  First part of this work will involve moving the `activeAccount` value to a URL prop so we can use the public ledger APIs to display information about accounts without requiring an actively connected hardware devise.


saminakh (approved):
So much work! Thank you for your hard work scott and perry 🎉

PR #486, opened 2022-04-14
feat: Activate accounts just before a protected action

staylorwr:


P-v-R (approved):
Nice work. looks good to me

PR #490, opened 2022-04-19
feat: Add a new ledger device and add a new ledger account

staylorwr:
https://www.loom.com/share/3747fa921497480dacbdc4c677928e4a

### DIFF
--- a/src/composables/useWallet.ts
+++ b/src/composables/useWallet.ts
@@ -23,17 +23,13 @@ import { Router } from 'vue-router'
 import { Subscription, Subject, firstValueFrom, zip } from 'rxjs'
 import { touchKeystore, hasKeystore, initWallet as createNewWallet, storePin } from '@/actions/vue/create-wallet'
 import {
-  deleteHardwareWalletAddress,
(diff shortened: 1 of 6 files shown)


src/composables/useWallet.ts
```diff
@@ -278,6 +261,70 @@ const accountNameFor = (accountAddress: AccountAddressT): string => {
...
+    hardwareWalletConnection: HardwareWalletLedger.create({
+      send: sendAPDU
+    }),
+    alsoSwitchTo: false,
+    verificationPrompt: false
+  }
+  try {
+    const wallet = await firstValueFrom(radix.__wallet)
+    const connectedDeviceAccount = await firstValueFrom(wallet.deriveHWAccount(data))
+    const hardwareDevice = hardwareDevices.value.find((hw) => hw.addresses.find((addr) => addr.address.equals(connectedDeviceAccount.address)))
```
P-v-R, line 282: This appears to be returning `undefined `for me, resulting in the index stored in `wallet.json `to be 0 and bypassing the below `if (hardwareDevice) {...}` block

PR #496, opened 2022-04-22
Fix Latest builds by bumping the version

staylorwr:
The store migrations won't run on latest because it's still locked to the same version.


saminakh (approved):
💥

PR #501, opened 2022-05-03
fix: Update wallet creation and restore for new routing

staylorwr:
Updates the create and restore wallet flows for the new route-driven active accounts


src/composables/useWallet.ts
```diff
@@ -250,12 +250,14 @@ const switchAddress = (address: AccountAddressT) => {
   }
 }
 
-const setActiveAddress = (accountAddressValue: string) => {
+const setActiveAddress = async (accountAddressValue: string) => {
   const address = AccountAddress.fromUnsafe(accountAddressValue)
   if (address.isErr()) {
+    console.log(accountAddressValue)
```
P-v-R, line 256: @staylorwr Do we want to keep this log?

src/views/CreateWallet/index.vue
```diff
@@ -160,14 +160,18 @@ const CreateWallet = defineComponent({
       setWallet(newWallet.value)
       setPin(pin)
       loginWithWallet(passcode.value).then((client) => {
-        return firstValueFrom(client.ledger.networkId())
-      }).then((network) => {
+        return Promise.all([
+          firstValueFrom(client.ledger.networkId()),
+          firstValueFrom(client.activeAccount)
+        ])
+        // return firstValueFrom(client.ledger.networkId())
```
P-v-R, line 167: Seems like this comment may be unneeded

PR #504, opened 2022-05-18
feat: better update experience

P-v-R:
**WIP** - could use a second set of eyes on the code. I believe in order to test an update repeatedly and locally you need to change configuration variables that keep electron from following through with the download. but I could be wrong on that. [See loom video here](https://www.loom.com/share/7f210459e51f4bb2ac05e81c98cf6808)

To replicate update for testing on local environment 
* Set `version` in `package.json` to an earlier version than the current release IE `1.3.1`
* In `/background.ts` set `devTools: true`
* In `/vue.config` comment out `afterSign: ‘scripts/notarize.js’/`

**To Run Update**
* run `yarn run electron:build`
* once build complete, `open .` to open build in finder
* open file named `Radix Wallet-1.3.1-arm64.dmg`
* when given the prompt, choose to `replace`
	*`An older item named “Radix Wallet” already exists in this location. Do you want to replace it with the newer one you’re moving?`
* open your radix desktop wallet (not your code on a local server, but the actual desktop app) and it will be a copy of the code built locally. 

### 📸📸📸📸📸📸

Screen Shot 2022-05-18 at 7 01 43 AM
load


staylorwr (approved):
Code looks very sane! 👏🏼

PR #507, opened 2022-05-25
chore: Upgrade Electron and Electron Latest for newer tooling

staylorwr:
This PR simply updates electron & electron-builder so that builds no longer break, while we jump from a very early v12 release of electron to the latest of v13, there are very few breaking changes.


P-v-R (approved):
Do we need to include the `yarn.lock` file in this ? Looks good otherwise

staylorwr:
> Do we need to include the `yarn.lock` file in this ? Looks good otherwise

Yep @P-v-R, I need the lock file updates because I updated the dependencies.

PR #508, opened 2022-05-25
chore: notarize needs the team id

staylorwr:
Screen Shot 2022-05-25 at 5 56 38 PM
Screen Shot 2022-05-25 at 5 57 16 PM

Validated locally that this change correctly allows the app to notarize again.


P-v-R (approved):
Looks good to me, will reach out for all new envs for natasha, samina and I.

PR #514, opened 2022-05-31
fix: multiple error pop ups when the ledger

P-v-R:
Removed any rendering of the original boilerplate ledger error modal and am now rendering a prettier modal when a ledger-specific action results in an error. 

error


staylorwr (commented):
👏🏼

staylorwr (approved):

PR #515, opened 2022-06-01
fix: tokenBalances out of sync, not selecting native token

staylorwr:
https://www.loom.com/share/c92b9e4ac7f443c896792674f3b8067b

Fixes a bug in the transaction ui where the currency picker and selected currency was not preferring the token balance for the native token and was out of sync between changing accounts.


caleboller (approved):

caleboller:
Looks good to me!

PR #516, opened 2022-06-01
Chore: Minor UI Tweaks

caleboller:
This PR just makes a couple small tweaks that came up during QA.

- Removes a double divider line when there are no hardware accounts.
- Slightly widens the account name labels to provide for longer titles.

Before:
Screen Shot 2022-06-01 at 9 33 46 AM

After:
Screen Shot 2022-06-01 at 10 00 08 AM


P-v-R (approved):
looks good to me!

PR #517, opened 2022-06-01
feat: add check for amount greater than max

artsycoder533:
This PR checks if the `unstake` amount the user typed is greater than the `max-unstake` amount available.  If so, a notification is displayed informing the user that the amount they typed was rounded down to the `max-unstake` amount available, since you cant unstake more than you have.  Before it only checked if the `minDifference` amount was <=`0.000001` which did not account for any negative values.  If the `minDifference` is negative that means the user tried to `unstake` more than they have available.  If `minDifference` is positive, then the user entered a number lower than the `max-unstake` amount and a notification should only be displayed if this value is <= `0.000001`.

[See Linear Issue RDX-397 Here](https://linear.app/township/issue/RDX-397/user-not-accurately-notified-of-unstake-rounding)

### Before

https://user-images.githubusercontent.com/83678228/171427547-f72d69f9-7ae9-434f-a55d-71a3fa8849a1.mp4

### After

https://user-images.githubusercontent.com/83678228/171439740-1d1c2295-c2d1-4e4c-81aa-413e36b5b4d6.mp4

In `WalletStaking.vue`

Inside the `compareToMaxUnstakeAmount` function I added another constraint to the conditional  to determine which notification will be displayed to the user if any.

```
if (minDifference <= 0.000001) {
  setMaxUnstakeNotificationOn()
  setMaxUnstakeOn()

  if (minDifference < 0) {
    setMaxUnstakeOverageNotifcationOn()
  } else {
    setMaxUnstakeOverageNotifcationOff()
  }
}
```

I added another ref `showMaxUnstakeOverageNotification` to control which notification message is conditionally rendered inside of the notification div

```
<span v-if="showMaxUnstakeOverageNotification" class=" text-xs text-rBlue flex-1 ">
  {{$t('staking.maxUnstakeModeOverageNotification')}}
</span>
<span v-else class=" text-xs text-rBlue flex-1 ">{{$t('staking.maxUnstakeModeEnabledNotification')}}</span>
```

Inside `index.ts`

I added another property `maxUnstakeModeOverageNotification` inside of the staking object and set it equal to the notification message that will be displayed to the user for I8n compatibility.

```
maxUnstakeModeOverageNotification: 'The value you entered is greater than the full unstakable amount, therefore, max mode has been automatically enabled on this entry.'
```

### DIFF
--- a/src/text/index.ts
+++ b/src/text/index.ts
@@ -223,7 +223,8 @@ const messages = {
       unstakeMaxDisclaimer: "You're unstaking all tokens from",
       maxUnstakeButton: 'max',
       maxUnstakeCapitalized: 'MAX',
-      maxUnstakeModeEnabledNotification: 'The value you entered is within 0.000001 of the full unstakable amount, therefore, max mode has been automatically enabled on this entry.'
+      maxUnstakeModeEnabledNotification: 'The value you entered is within 0.000001 of the full unstakable amount, therefore, max mode has been automatically enabled on this entry.',
+      maxUnstakeModeOverageNotification: 'The value you entered is greater than the full unstakable amount, therefore, max mode has been automatically enabled on this entry.'
     },
     confirmation: {
       transferFromLabel: 'Your address',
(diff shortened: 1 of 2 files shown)


staylorwr (approved):
👍🏼 another great PR @artsycoder533

PR #520, opened 2022-06-02
chore: improve new ledger device pop up

P-v-R:
The old alert pop-up was being clipped by the accounts list side nav and wasn't styled like a _speech bubble_. This fix refactors the pop up to look more like the figma styling. 

### 📸📸📸📸
Screen Shot 2022-06-02 at 6 43 27 AM
popup

### DIFF
--- a/src/views/Wallet/WalletNewDevicePopup.vue
+++ b/src/views/Wallet/WalletNewDevicePopup.vue
@@ -1,10 +1,15 @@
 <template>
-  <div class="fixed w-screen h-screen z-20 flex items-center justify-center">
-    <div class="bg-white rounded-md w-72 absolute top-2/3 left-1/4 transform -translate-x-1/3 -translate-y-1/2 border border-rBlue bg-rGrayLight">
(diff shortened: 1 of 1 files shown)


artsycoder533:
This looks good! I don't have a hardware device yet to test this one out directly, but I did see the screenshot Katy posted on Slack showing how the alert was cut off.  I see you made it position absolute, so im assuming if the users screen was full of transactions it would overlap them and not cause a scroll.

P-v-R:
> This looks good! I don't have a hardware device yet to test this one out directly, but I did see the screenshot Katy posted on Slack showing how the alert was cut off. I see you made it position absolute, so im assuming if the users screen was full of transactions it would overlap them and not cause a scroll.

@artsycoder533 Yes, this pop-up will overlap the body of the radix app and not interfere with the behavior of the side nav, A user can scroll the side nav while the pop-up is visible and fixed to one location.

caleboller (approved):

PR #524, opened 2022-06-02
fix: stops user from pressing max without entering a validator address

artsycoder533:
This PR stops the user from pressing the `max` button if they have not first entered a `validator address`.  Previously the user could press the `max` button at any time which triggers the state of the `Request Unstake` button to change from disabled to active. As a result, the user can then click the `Request Unstake` button which causes an error, but the user isnt aware, because it doesnt provide any feedback to the user.  Now the state of the `max` button is controlled based on whether the user has entered a value in the `validator address`.  Once an address is entered, the disabled styling of the `max` button is removed.

[See Linear Issue RDX-400 Here](https://linear.app/township/issue/RDX-400/max-unstake-button-should-be-disabled-if-no-validator-address)

In `WalletStaking.vue`

I used class binding to switch the styling of the `max` button to match the disabled `Request Unstake` button only if `formValidatorName` does not have a value.  Once `formValidatorName` has a value, this stying is not applied.

```
<button
  @click.prevent="setMaxUnstakeOn"
  v-if="activeForm == 'UNSTAKING'"
  class="rounded border border-rGreen text-rGreen w-2/12 h-full ml-6"
  :class="formValidatorName === '' && 'bg-rGray text-rGrayDark cursor-not-allowed border border-rGray'"
  :disabled="formValidatorName === ''"
>
  {{ $t('staking.maxUnstakeButton') }}
</button>
```

### Before

https://user-images.githubusercontent.com/83678228/171727953-7633b982-986a-41cd-a0a8-8a99e9928d8b.mp4

### After

https://user-images.githubusercontent.com/83678228/171890783-97686cf1-4b7c-4948-8215-b3d98bd97704.mp4

### DIFF
--- a/src/views/Wallet/WalletStaking.vue
+++ b/src/views/Wallet/WalletStaking.vue
@@ -476,7 +476,6 @@ const WalletStaking = defineComponent({
     })
 
     const emptyFormValidatorName: ComputedRef<boolean> = computed(() => {
-      console.log(!!formValidatorName.value)
       return !formValidatorName.value
     })
 
--- a/src/views/Wallet/WalletStaking.vue
+++ b/src/views/Wallet/WalletStaking.vue
@@ -90,8 +90,7 @@
                   <button
                     @click.prevent="setMaxUnstakeOn"
                     v-if="activeForm == 'UNSTAKING'"
-                    class="rounded border border-rGreen text-rGreen w-2/12 h-full ml-6"
-                    :class="formValidatorName === '' && 'bg-rGray text-rGrayDark cursor-not-allowed border border-rGray'"
+                    :class="{'rounded border border-rGreen text-rGreen w-2/12 h-full ml-6': true, 'bg-rGray text-rGrayDark cursor-not-allowed border border-rGray': formValidatorName === ''}"
                     :disabled="formValidatorName === ''"
                   >
                     {{ $t('staking.maxUnstakeButton') }}
--- a/src/views/Wallet/WalletStaking.vue
+++ b/src/views/Wallet/WalletStaking.vue
@@ -51,6 +51,7 @@
               :placeholder="$t('staking.validatorPlaceholder')"
               rules="required|validValidator"
               :validateOnInput="true"
+              ref="address"
             />
             <FormErrorMessage name="validator" class="text-sm text-red-400" />
           </div>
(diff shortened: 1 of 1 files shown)


src/views/Wallet/WalletStaking.vue
```diff
@@ -91,6 +91,8 @@
                     @click.prevent="setMaxUnstakeOn"
                     v-if="activeForm == 'UNSTAKING'"
                     class="rounded border border-rGreen text-rGreen w-2/12 h-full ml-6"
+                    :class="formValidatorName === '' && 'bg-rGray text-rGrayDark cursor-not-allowed border border-rGray'"
```
staylorwr, line 94: Just one style nit on the code structure, everything else is looking great!

You can use the `:class` and _object binding_ to define a conditional object of class names, it removes the dual-definition of `class`

```
                    :class="{
                      'rounded border border-rGreen text-rGreen w-2/12 h-full ml-6': true,
                      'bg-rGray text-rGrayDark cursor-not-allowed border border-rGray': formValidatorName === ''
                    }"
```

Since you're calling this `formValidatorName` twice in a few lines and it's already a computed property, you should also be able to write a new computed property to remove that duplication / be clearer with the intent:

```js
    const formValidatorName: ComputedRef<string | undefined> = computed(() => {
      if (!values.validator) return ''
      if (!safeAddress) return ''
      const v = maybeGetValidator(safeAddress)
      return v ? v.name : ''
    })

  const emptyValidatorName: ComputedRef<boolean> = computed(() => {
     return !!formValidatorName.value 
  })
```

```suggestion
                    :class="{
                      'rounded border border-rGreen text-rGreen w-2/12 h-full ml-6': true,
                      'bg-rGray text-rGrayDark cursor-not-allowed border border-rGray': emptyValidatorName
                    }"
```

staylorwr (changes requested):
Just one little style nit, everything else looks great!

src/views/Wallet/WalletStaking.vue
artsycoder533: I implemented the suggested changes but I made one small change, the double bang operator inside of the `emptyValidatorName` computed property gave the opposite values than what I wanted so, I changed this to a single not and it works the same as before.

```
const emptyFormValidatorName: ComputedRef<boolean> = computed(() => {
      return !formValidatorName.value
    })
```

staylorwr (approved):

PR #527, opened 2022-06-06
Verify Address and Initial HW wallet bugs

staylorwr:
https://www.loom.com/share/f9e3042553144e07af3aceb057ee5659

Depends on the changes in https://github.com/radixdlt/radixdlt-javascript/pull/202

This PR cleans up two edge cases where the SDK / Wallet were depending on `wallet` state to correctly derive and message the correct ledger device.

**Verify HW Address**
From my understanding of the SDK, in the above PR, I've updated the wallet functionality for displaying the address on the ledger device to no longer observe and pipe through the `activeAccount` observable.  On major wallet / hw account / active account changes, this caused the verify function to send three+ equivalent, duplicate messages through the `sendAPDU` tunnel.  

**Deriving new HW Devices**
It seems that another side effect of this observable magic is that hw accounts would not always correctly derive.

### DIFF
--- a/src/actions/electron/hardware-wallet.ts
+++ b/src/actions/electron/hardware-wallet.ts
@@ -1,20 +1,27 @@
 import TransportNodeHid from '@ledgerhq/hw-transport-node-hid'
 import { IpcMainInvokeEvent } from 'electron/main'
 
-export const sendAPDU = async (event: IpcMainInvokeEvent, apdu: { cla: number, ins: number, p1: number, p2: number, data?: string }) => {
+type APDU = { cla: number, ins: number, p1: number, p2: number, data?: string }
+let lastAPDU: APDU | null = null 
+
+export const sendAPDU = async (event: IpcMainInvokeEvent, apdu: APDU) => {
   const paths = await TransportNodeHid.list()
   if (paths.length === 0) throw Error('No device found.')
-
-  const transport = await TransportNodeHid.open(paths[0])
+  if (paths.length > 1) throw Error('Too Many Devices Enabled')
+  if (lastAPDU && lastAPDU === apdu) {
+    console.warn('duplicate message', apdu)
+    return
+  }
   let result
   try {
+    const transport = await TransportNodeHid.open(paths[0])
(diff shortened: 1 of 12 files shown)


src/actions/electron/hardware-wallet.ts
```diff
@@ -1,20 +1,27 @@
...
+  }
   let result
   try {
+    const transport = await TransportNodeHid.open(paths[0])
     result = await transport.send(apdu.cla, apdu.ins, apdu.p1, apdu.p2, apdu.data ? Buffer.from(apdu.data, 'hex') : undefined)
-  } catch (e) {
+    lastAPDU = apdu
     await transport.close()
+  } catch (e) {
+    console.log('hi there', e)
```
P-v-R, line 22: hello! assuming we don't want this console.log?

src/actions/electron/hardware-wallet.ts
```diff
@@ -1,20 +1,27 @@
...
+let lastAPDU: APDU | null = null 
+
+export const sendAPDU = async (event: IpcMainInvokeEvent, apdu: APDU) => {
   const paths = await TransportNodeHid.list()
   if (paths.length === 0) throw Error('No device found.')
-
-  const transport = await TransportNodeHid.open(paths[0])
+  if (paths.length > 1) throw Error('Too Many Devices Enabled')
+  if (lastAPDU && lastAPDU === apdu) {
+    console.warn('duplicate message', apdu)
```
P-v-R, line 12: Do we want to keep this console warn in as well as log below?

P-v-R (changes requested):

artsycoder533 (commented):
I am now able to connect both of my ledgers! When I press on the click to copy button a pop up is shown that asks to verify the address matches whats currently shown on my ledger.  Both addresses match. Im not sure if this if this means anything, but the same time I press the click to copy button an error is displayed in the console when the verify address pop appears with a number 3 beside the error.
Screen Shot 2022-06-07 at 10 45 04 AM
Screen Shot 2022-06-07 at 11 10 52 AM

P-v-R (approved):

PR #528, opened 2022-06-07
chore: improve new device pop up bttn hover styling

P-v-R:
The on-hover styling of the "got it" button was only changing the color of the button text and not the whole button, this fixes that. Also tweaked the sizing of it. 

### 📸 📸 📸 📸 📸
popuip


artsycoder533 (commented):
I saw the before of this in Scotts PR, but when I pull down your branch and click on 'add hardware account'  I get the 'Deriving Account' pop up with the spin wheel and then the new account is just there, the modal in the bottom left doesn't shop up anymore for me.

caleboller (approved):
Looks good to me!

PR #531, opened 2022-06-15
fix: Unable to encrypt multiple transactions with ledger

staylorwr:
This PR refactors the `useWallet` and `useTransactions` APIs by removing the callback from `activateAccount` and instead return a promise that includes both the newly activated account and a reference to the client that has the account activated.  This is used in the transaction build process to correctly reference both the valid active account and the activated client.  
Previously there was a bug where the client and sdk were not correctly in sync, causing a validation error deep in the sdk when a ledger was used to encrypt more than one transaction in a row.

In older versions of the app, we relied on a `hardwareAddress: string` from the electron store as a reference to the single hardware account that was associated on a ledger device.  This PR also cleans up that reference and replaces it with to the list of `hardwareDevices` from the electron store.  The previous value was not always correctly set and led to strange verify behavior.


artsycoder533 (commented):
I was able to send tokens between two accounts on the same ledger device 10 times without any errors or getting stuck on the confirmation screen.  I repeated this same process for both of my ledgers.  The only thing I noticed which im not sure if its related to the changes made or not, but the history seems to be cycling through between 3 different accounts.  Its doing this much more frequently than in the original video from Kate, its almost instant.  It got to a point it was hard to find my test transactions because it cycled through so much.  See video.

https://user-images.githubusercontent.com/83678228/173886605-17cf465c-bcff-40ec-96e3-95211ec7ea1f.mov

artsycoder533 (approved):

PR #533, opened 2022-06-16
Refactor: Unify `radix` state.

staylorwr:
This commit goes through the hard work of consolidating any references
to the main `radix` sdk into a single `useWallet`.  There have been
problems with inconsistent ledger behavior while processing transactions
of all forms while swapping between active accounts and physical ledger
devices.

Originally this larger scope of work had been built on the premise that
the radix sdk must be `activated` with the referenced active account
just before a required action was performed.  It seems as though the
SDK's state does not transfer around between references, callback and
promises; nor should we expect it to.

This commit unifes those interactions by:

- removing all references to `activateAccount` from outside of the
  `useWallet` API.
- refactor the behavior of `useTransactions` into the main `useWallet`
  API.  This allows us to only reference the main SDK, and no longer
  depend on intermediate references that seem to eventually fall out of
  sync.
- refactor `useHistory` to move decryption of messages back to the
  `useWallet` sdk.  `useHistory` still serves as a holder of decrypted
  messages.
- Since `activateAccount` is now internal to `useWallet`, call it before
  `transfer`, `stake` or `unstake` actions.  `activateAccount` now acts
  only on the main `radix` instance and we can no longer attempt to
  inject new Wallets into the SDK (I'd guess this was the source of the
  bugs)
- Guard against double clicks on long-running account actions by adding
  a `isActivating` boolean that discards activation attempts while an
  activation is ongoing


artsycoder533 (changes requested):
I completed 3 test runs of 30 transactions each sending tokens between two accounts on the same ledger. Two rounds were with one ledger, and one round with the second ledger. Here are the results:

### Ledger 1:
`Round 1`, 1 hex error / 30. 
After confirming the first encryption prompt on the ledger just before the entering pin screen would have shown.
Screen Shot 2022-06-16 at 3 54 35 PM
Screen Shot 2022-06-16 at 4 07 26 PM

`Round 2` 2 hex errors / 30. 
The first error occured in the same spot as the previous round, while the second error occurred after the second round of confirm encryption prompts.
Screen Shot 2022-06-17 at 8 58 08 AM
Screen Shot 2022-06-17 at 9 17 58 AM

### Ledger 2:
`Round 1`, 5 hex errors/30 
3 at the first confirm encryption prompt, 2 at the second confirm encryption prompt
Screen Shot 2022-06-17 at 10 09 02 AM

Im assuming the `hex error` is referring to the encrypted message itself? It seems like when the user has to confirm an encryption, something is sent and used as maybe a parameter somewhere else? If its not the right type it throws an error? Im wondering whatever the next step is in the process after a user confirms an encryption could there be a check in place to see if the hex is the right type.  If not, on the enter your pin screen I believe theres a `cancel transaction` button, I wonder could that be tied into so at the least the user isnt stuck if the hex isnt the right type, the transaction is just cancelled and they can try it again, or is it too late at that point since the transaction has already been confirmed? Thinking out loud but it seems like it may be easier to find a solution for when this error is thrown rather than trying to not get any errors at all?  The first ledger  2 errors max wasnt bad, but that second ledger was not playing fair!

PR #538, opened 2022-06-23
fix: ensure the correct account is activated before moving on

staylorwr:


src/composables/useWallet.ts
```diff
@@ -613,42 +613,35 @@ const createNewHardwareAccount = async () => {
...
+  hardwareInteractionState.value = 'DERIVING'
+  const hwAccount: AccountT = await firstValueFrom(radix.deriveHWAccount({
+    keyDerivation: HDPathRadix.create({
+      address: { index: hwaddr.index, isHardened: true }
+    }),
+    hardwareWalletConnection,
+    alsoSwitchTo: true,
+    verificationPrompt: false
+  }))
+  console.log('connected to ', hwAccount.address.toString())
```
P-v-R, line 627: this is the only ugly thing I see, unless we want???

P-v-R (approved):

PR #545, opened 2022-07-20
fix: Remove undefined functions

staylorwr:
Latest has a regression where we're missing two files, causing the background.ts boot to fail.


artsycoder533 (approved):
I was able to load the app and toggle between us/european decimal types.

PR #553, opened 2022-08-23
fix: Watch active address query param in settings

staylorwr:


artsycoder533 (approved):
selected account now highlighted when in settings

PR #558, opened 2022-08-23
fix: New Ledgers should be named Ledgers

staylorwr:
[Linear Issue RDX-437](https://linear.app/township/issue/RDX-437)


artsycoder533 (approved):
Re-added 2 ledgers and they're now named Ledger 1 & Ledger 2.

PR #562, opened 2022-08-25
chore: Update and align sidenav

staylorwr:


artsycoder533 (approved):
this seems to fix the issues with ticket 445 and 439

PR #564, opened 2022-08-26
chore: add confirm message encryption modal

artsycoder533:
This PR adds a confirm message encryption modal as a first step when initiating a ledger transaction with an encrypted message.

[See Linear Ticket RDX-443 here](https://linear.app/township/issue/RDX-443)


P-v-R (approved):
nice work

saminakh (approved):
👍

PR #567, opened 2022-08-30
fix: Update SDK for message encryption patch

staylorwr:
Upgrades the SDK for the fix in radixdlt/radixdlt-javascript#208 which updates
the message encryption shared secret calculation so that the buffer is always
the correct length.


artsycoder533 (approved):
Completed 31 test transactions with no hex errors. I also let the ledger go idle and resumed transactions.

PR #569, opened 2022-08-31
chore: add condition to decrypting modal body copy

artsycoder533:
This PR adds a conditional to the body of the 'Decrypting Message' modal that differentiates between a software and hardware account, so the the body is only shown when the active account is a hardware account.

Update:
I renamed the 'isSoftwareAccount' function to 'activeAddresIsSoftwareAccount' per Scott's suggestion and ensured the 'Submitting Transactions' modal is only shown at the end of a send tokens transaction, 

[See Linear Ticket RDX-448](https://linear.app/township/issue/RDX-448)

### DIFF
--- a/src/composables/useWallet.ts
+++ b/src/composables/useWallet.ts
@@ -105,11 +105,9 @@ const wallet: Ref<WalletT | null> = ref(null)
 const latestAddress: Ref<string> = ref('')
 const loadingLatestAddress: Ref<boolean> = ref(true)
 const transactionSubs = new Subscription()
-
 const activeMessage: Ref<string> = ref('')
 const activeMessageInTransaction: Ref<MessageInTransaction | null> = ref(null)
 const activeTransactionForm: Ref<string | null> = ref(null)
-
 const selectedCurrency: Ref<Decoded.TokenAmount | null> = ref(null)
 const shouldShowConfirmation: Ref<boolean> = ref(false)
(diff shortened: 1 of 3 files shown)


src/composables/useWallet.ts
```diff
@@ -119,6 +119,19 @@ const unstakeInput: Ref<UnstakeTokensInput | null> = ref(null)
 const transactionErrorMessage: Ref<string | null> = ref(null)
 const transactionFee: Ref<AmountT | null> = ref(null)
 const transferInput: Ref<TransferTokensInput | null> = ref(null)
+const showLedgerInteractionModalBody: Ref<boolean> = ref(true)
+
+// check if software or hardware address
+const isSoftwareAccount = () => {
```
staylorwr, line 125: Can we relabel this to `activeAddresIsSoftwareAccount`?  This name makes me assume that it would accept an `Address` as an argument.

staylorwr (changes requested):
Thanks for this @artsycoder533, can we update the name of the new function to be more explicit?

src/composables/useWallet.ts
artsycoder533: @staylorwr  Sure thing, Caleb asked me to get rid of the 'Decrypting Account' modal too when sending messages from a Software Account so i'll be adding to this PR.

staylorwr (approved):

PR #573, opened 2022-08-31
fix: removing validator address doesn't disable 'Unstake' button

artsycoder533:
This PR adds an onChange handler 'checkForEmptyValidator' that checks if the user changed the validator input. Previously the user could fill both the validator and amount in, then if the validator address was removed, the 'Request Unstake' button did not go back to a disabled state.  As a result, the user could click this button and nothing would happen.

[See Linear Ticket RDX-450 Here](https://linear.app/township/issue/RDX-450)

### DIFF
--- a/src/views/Wallet/WalletStaking.vue
+++ b/src/views/Wallet/WalletStaking.vue
@@ -350,7 +350,6 @@ const WalletStaking = defineComponent({
     }
 
     const checkForEmptyValidator = () => {
-      // if user removes validator address
       const validator = safelyUnwrapValidator(values.validator)
       validateField('validator')
       if (!validator) {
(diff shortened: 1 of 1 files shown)


src/views/Wallet/WalletStaking.vue
```diff
@@ -465,6 +476,7 @@ const WalletStaking = defineComponent({
       const safeAddress = safelyUnwrapValidator(values.validator)
       if (!safeAddress) return ''
       const v = maybeGetValidator(safeAddress)
+      console.log(v, v?.name)
```
staylorwr, line 479: can you remove the console log?

staylorwr (changes requested):

caleboller (approved):

staylorwr (approved):

PR #583, opened 2022-11-15
fix(styles): Realign sidenav icons between hardware & software wallet accounts

staylorwr:


artsycoder533 (approved):
@staylorwr Looks good to me!  Sorry I forgot about this one!

PR #585, opened 2022-12-01
Natasha/rdx 464 bug issue when restoring wallet from

artsycoder533:
This PR fixes inputs in the `restore wallet from seed phrase` and `change pin` screens not allowing a user to type after pressing the `-` key.  The input was changed from type=number to type=text, then a regex was used to find and replace the `e` with an empty space.  When a user clicks in the `enter pin` input it now clears both inputs and allows the user to type again.  The error message copy was also updated.

[Linear Ticket RDX-464](https://linear.app/township/issue/RDX-464)

### DIFF
--- a/src/views/CreateWallet/CreateWalletCreatePin.vue
+++ b/src/views/CreateWallet/CreateWalletCreatePin.vue
@@ -5,9 +5,10 @@
       <pin-input
         name="pin"
         :values="values.pin"
-        :autofocus="updatingFirstInput"
+        :autofocus="activePin === 1"
         :large="true"
-        @finished="handleFinished"
+        @click="firstPinClicked"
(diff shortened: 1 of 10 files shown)


src/views/CreateWallet/CreateWalletCreatePin.vue
```diff
@@ -47,52 +53,71 @@ const CreateWalletCreatePin = defineComponent({
...
-    return {
-      errors,
-      values,
-      meta,
-      resetForm,
-      setErrors
+    const activePin = ref(0)
+    const isValidPin = ref(false)
+
+    const resetFormForNonmatchingPins = () => {
```
artsycoder533, line 63: clear the inputs after a mismatch 👌🏾

PR #594, opened 2023-03-07
feat: Export Wallet

staylorwr:


src/views/Settings/SettingsExport.vue
```diff
@@ -0,0 +1,308 @@
...
+      const localAddrs = localAccounts.value.map((account) => account.address.toString())
+      const hardwareAddresses = hardwareDevices.value.flatMap((device) => device.addresses.map((address) => address.address.toString()))
+      selectedAccounts.value = [...localAddrs, ...hardwareAddresses]
+    }
+
+    const accountSummary = (account: AccountT, isLocal: boolean): string => {
+      const name = accountNameFor(account.address)
+      completedExports.value = completedExports.value + 1
+      const localType = isLocal ? 'S' : 'H'
+      const addressIndex = account.hdPath?.addressIndex.index || 0
```
CyonAlexRDX, line 215: I don't think we care about non-HD accounts, since I don't think Olympia wallet supports it. So best to not have this fallback (` || 0`), which is semantically wrong. Instead assert that `hdPath` is not null and before that, part of `localAccounts.value.filter` below, filter out accounts which do not contain `hdPath` (they all should).

src/views/Settings/SettingsExport.vue
```diff
@@ -0,0 +1,308 @@
...
+      return accountToExportPayload(localType, account.publicKey.toString(true), addressIndex, name)
```
CyonAlexRDX, line 216: furthermore, you must "unharden" the `addressIndex`, sorry for not specifying that, so if `addressIndex` is larger than `hardenedIncrement` (defined here https://github.com/radixdlt/radixdlt-javascript/blob/main/packages/crypto/src/elliptic-curve/hd/bip32/bip32PathComponent.ts#L5, as hexadecimal `0x80000000`) you need to subtract `hardenedIncrement`, and this is true for all `addressIndex` by default. i.,e. for first account the value of you export should not be `2147483648` - which is **_decimal_** for `0x80000000` - but rather 0. because 2147483648 - 2147483648 = 0 (or 0x80000000 - 0x80000000 = 0). And analogously the value for the next account `2147483649` should be `1` because 2147483649 - 0x80000000 = 1. etc etc

This makes the export format much more compact, since we save 9 characters per account 😃

CyonAlexRDX, line 216: furthermore you are incorrectly calling `publickey.toString(true)`, that is wrong, that is resulting in hexadecimal strings. You should Base64 encode the public key compressed data.

tests/unit/helpers/exportAsCode.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+})
+
+describe('exportAsCode', () => {
+  it('correctly exports all accounts and adds headers to payloads', () => {
+    testCases.forEach(({ olympiaWallet: { accounts, mnemonic }, payloadSizeThreshold, numberOfPayloads, payloads }) => {
+      const data = testAccountsToExports(accounts)
+      const mnemonicLength = mnemonic.split(' ').length
+      const result = exportAsCode(data, payloadSizeThreshold, mnemonicLength)
+      expect(result.length).to.equal(numberOfPayloads)
+      expect(result).to.deep.equal(payloads)
```
CyonAlexRDX, line 26: this deep equals check incorrectly thought the base64 encoded publickeys and hex encoded public keys were equal, which is true on a bytearray level, but since the export format is **string** based, rather each string representation of each payload must match the expected one.

src/helpers/exportAsCode.ts
```diff
@@ -0,0 +1,27 @@
+const END__OF_HEADER = ']'
+const INTRA_SEPARATOR = '^'
+const INTER_SEPARATOR = '~'
+const END_OF_ACCOUNT_NAME = '}'
+
+const ACCOUNT_NAME_REPLACEMENT = '_'
+
+type AccountType = 'H' | 'S'
+
+export const accountToExportPayload = (accountType: AccountType, publicKey: string, addressIndex: number, name: string): string => {
```
CyonAlexRDX, line 10: publicKey here should be Base64 encoding of the compressed public key bytes, currently you are incorretly hex encoding the key.

src/helpers/exportAsCode.ts
```diff
@@ -0,0 +1,27 @@
...
+  const payloadsWithoutHeaders = allAccounts.match(new RegExp(`.{1,${payloadSize}}`, 'g')) || []
+  const payloadTotal = payloadsWithoutHeaders.length
+  return payloadsWithoutHeaders.map((payload, index) => {
+    return `${payloadTotal}${INTRA_SEPARATOR}${index}${INTRA_SEPARATOR}${mnemonicLength}${END__OF_HEADER}${payload}`
+  })
+}
+
+export const sanitizeName = (name: string): string => {
+  const limitedName = name.substring(0, 30)
+  const replacedName = limitedName.replace(/[\]^~}]/, ACCOUNT_NAME_REPLACEMENT)
```
CyonAlexRDX, line 25: this replacement is wrong, you are not using the correct characters, rather 
```javascript
var limitedName = name.substring(0, 30)
for forbidden in [END__OF_HEADER, INTRA_SEPARATOR, INTER_SEPARATOR, END_OF_ACCOUNT_NAME] {
  limitedName = limitedName.replace(forbidden, ACCOUNT_NAME_REPLACEMENT)
}
```
or similar. 

Because I just did an export which contained both `^` and `~` using build https://github.com/radixdlt/olympia-wallet/actions/runs/4766606231

CyonAlexRDX (changes requested):
Almost there! Three issue (I tried this build: https://github.com/radixdlt/olympia-wallet/actions/runs/4766606231):
1) you are not base64 encoding the compresed public key, instead you are doing hex.
2) you should use non hardened address index, this was a miss on my part not specifying that
3) some forbidden chars slipped through in account name it seems

For Base64 encoding, I'm reposting what I've specified in the format:

> The public key is the 33 bytes compressed public key base64 encoded. IMPORTANT! Notice the difference of base64 encoding the public key as utf8 string, vs encoding a Buffer (hex), we should not base64 the public key as an utf8 string. Test this online tool: [Base64 Encode Online](https://www.rapidtables.com/web/tools/base64-encode.html) and Base64 this 02f669a43024d90fde69351ccc53022c2f86708d9b3c42693640733c5778235da5 public key first as Input Type = Text String which yields this ❌ "MDJmNjY5YTQzMDI0ZDkwZmRlNjkzNTFjY2M1MzAyMmMyZjg2NzA4ZDliM2M0MjY5MzY0MDczM2M1Nzc4MjM1ZGE1" ❌ long string, then change the Input Type = Hex, and encode and you get ✅ "AvZppDAk2Q/eaTUczFMCLC+GcI2bPEJpNkBzPFd4I12l" ✅. So correct Base64 encoding decreases 66 chars hex to 44 chars base64, whereas incorrect encoding increases 66 chars to 88 chars.

staylorwr:
Should be all cleaned up @CyonAlexRDX Thanks for the thorough review 🙆🏼‍♂️

CyonAlexRDX (approved):
LGTM!

PR #597, opened 2023-07-11
fix: Sanitize account names while displaying them for export

staylorwr:
Account names can contain a lot of characters that need to be sanitized before exporting them to the Babylon Wallet.  This PR updates the export interface to display the sanitized names, so that it's easier to compare actual export results compared to expectations.

### DIFF
--- a/src/views/Settings/ExportAccountListItem.vue
+++ b/src/views/Settings/ExportAccountListItem.vue
@@ -11,7 +11,7 @@
     </div>
     <div class="flex flex-col gap-2 flex-1">
       <div>
-        <span v-if="name" class="text-sm mr-2">{{ name }}</span>
+        <span v-if="name" class="text-sm mr-2">{{ cleanedUpName }}</span>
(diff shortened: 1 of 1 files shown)


src/views/Settings/ExportAccountListItem.vue
```diff
@@ -48,9 +49,16 @@ export default defineComponent({
   setup (props) {
     const address = toRef(props, 'address')
     const selected = toRef(props, 'selected')
+    const name = toRef(props, 'name')
+    const cleanedUpName = computed(() => {
```
artsycoder533, line 53: 👍🏾 I had to go back to the helper folder and find that sanitize function to see what it was doing behind the scenes.  Looks like it takes the account name and replaces any emojis or unicode, chops that string off at 30 characters , removes any forbidden characters, and adds that end of account constant to the end of it then returns all that as the new name.

artsycoder533 (approved):
Looks sane to me!

PR #598, opened 2023-07-12
feat: Delete Wallet Settings Tab

staylorwr:
### Overview

Add a 'Delete Wallet' tab to Settings that allows a user to clear their wallet.

The settings tab includes a 'Delete All Wallet Data' button that redirects
the user to the 'Delete Wallet' part of the 'Forgot My Password' flow.


artsycoder533 (approved):
When I click the 'Delete All Wallet Data' in settings I get taken to the same screen as if i had chosen 'forgot my password' link on the login screen followed by the red 'delete and restore from seed phrase ' button.  According to the slack thread that sounds like what Matt wants!

